PR #47, opened 2022-10-05
support ConstExprNode node to avoid crashing with 500

ildyria:
This allows the AST parser to not break on a ConstExprNode when going through documentation.

2022-10-05_1920x1080_12:05:04

### DIFF
--- a/src/PhpDoc/PhpDocTypeWalker.php
+++ b/src/PhpDoc/PhpDocTypeWalker.php
@@ -2,6 +2,7 @@
 
 namespace Dedoc\Scramble\PhpDoc;
 
+use PHPStan\PhpDocParser\Ast\ConstExpr\ConstExprNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayShapeItemNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayShapeNode;
 use PHPStan\PhpDocParser\Ast\Type\ArrayTypeNode;
@@ -13,7 +14,7 @@ use PHPStan\PhpDocParser\Ast\Type\UnionTypeNode;
 
 class PhpDocTypeWalker
 {
-    public static function traverse(TypeNode $type, array $visitors)
+    public static function traverse(TypeNode|ConstExprNode $type, array $visitors)
     {
         $callVisitors = function ($node, $method) use ($visitors) {
             foreach ($visitors as $visitor) {


romalytvynenko:
@ildyria Thanks for the fix!

Can you please share an example of the PhpDoc that causes this fail?

Thanks!

ildyria:
> @ildyria Thanks for the fix!
> 
> Can you please share an example of the PhpDoc that causes this fail?
> 
> Thanks!

That is going to be difficult to find, I am using scramble as direct plug on https://github.com/LycheeOrg/Lychee and the project is quite big. I will try to find one.

ildyria:
Found it. This will crash.
```php
	/**
	 * Reset the token of the currently authenticated user.
	 *
	 * @return array{'token': string}
	 *
	 * @throws UnauthenticatedException
	 * @throws ModelDBException
	 * @throws \Exception
	 */
	public function resetToken(): array
{...}
```

Found it. This will pass.
```php
	/**
	 * Reset the token of the currently authenticated user.
	 *
	 * @return array
	 *
	 * @throws UnauthenticatedException
	 * @throws ModelDBException
	 * @throws \Exception
	 */
	public function resetToken(): array
{...}
```

Note that this is valid and follows https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

romalytvynenko:
I see. 

But as far as I can tell, suggested fix will result in loosing type information.

I believe that it should be a more comprehensive support of the const expressions in PhpDoc (which will result in a correct type).

So the PhpDoc comment:
```php
/**
 * @var $foo array{'token': string}
 */
```

Should become type:
```
array{token: string}
```

With the suggested fix it will become an `unknown` type key:
```
array{unknown: string}
```

ildyria:
Said function is:
```php
	/**
	 * Reset the token of the currently authenticated user.
	 *
	 * @return array{'token': string}
	 *
	 * @throws UnauthenticatedException
	 * @throws ModelDBException
	 * @throws \Exception
	 */
	public function resetToken(): array
	{
		/** @var User $user */
		$user = Auth::user() ?? throw new UnauthenticatedException();
		$token = strtr(base64_encode(random_bytes(16)), '+/', '-_');
		$user->token = hash('SHA512', $token);
		$user->save();

		return ['token' => $token];
	}
```

With this patch, the doc becomes:
2022-10-05_1920x1080_13:43:06

romalytvynenko:
Fixed in https://github.com/dedoc/scramble/pull/59

Issue #44, opened 2022-10-02
Default config file causes error


mortenscheel:
Steps to reproduce:

1. Install scramble in clean Jetstream project
2. Publish scramble config
3. Run `php artisan serve`
```
In UrlGenerator.php line 121:

  Illuminate\Routing\UrlGenerator::__construct(): Argument #2 ($request) must be of type Illuminate\Http\Request, null given, called in /Users/**redacted**/vendor/laravel/framew
  ork/src/Illuminate/Routing/RoutingServiceProvider.php on line 67
```
Commenting out the `'api_base_url'` line fixes it

romalytvynenko:
Hey @mortenscheel 

Thank you so much for reporting.

The issue was fixed in `0.5.0`: https://github.com/dedoc/scramble/releases/tag/v0.5.0

P.S.: Please read the release description as the fix is a breaking change and may affect you if you've changed `api_base_url`.

PR #67, opened 2022-10-29
Add option for multiple servers

woutervandamme:
Add the possibility of having multiple servers show in the server section.

Screenshot 2022-10-30 at 00 09 26

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -10,6 +10,13 @@ return [
      */
     'api_path' => 'api',
 
+    /*
+     * List the servers that are used in your API.
+     */
+    'servers' => [
+        'Live' => url(config('scramble.api_path', 'api')),
+    ],
+
     'info' => [
         /*
          * API version.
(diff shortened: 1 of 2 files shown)


romalytvynenko:
Hey @woutervandamme 

Sorry for the long time it took me to look into it.

This is a nice idea and I was thinking about how to implement it. There is an issue with using `url` in config file that causes Laravel app to throw an exception when using built-in server: https://github.com/dedoc/scramble/issues/44

Do you have an ideas how to implement this feature without referring to `url` in config? Or maybe there is a way to use `url` and not get an exception?

Thanks!

woutervandamme:
I will have some time next week, i'll take a look at it!

romalytvynenko:
Hey @woutervandamme 

I've added this feature here: https://github.com/dedoc/scramble/pull/76

Basically the implementation doesn't use `url/config` fns in servers config and this way it should guarantee safety. 

Thanks for your contribution!

PR #91, opened 2022-12-05
Use interface instead of reflections

ildyria:
Simply aims to make the code more robust :)


romalytvynenko:
Hey @ildyria 

I appreciate your time and effort but for now I'm not going to merge this change and will keep internal code structure as it is.

Thanks.

PR #102, opened 2023-01-13
Access to docs in `local` env only by default and using `viewApiDocs` gate in other environments

tontonsb:
Hey,

This is a bit opinionated and I am aware I can remove or replace the middleware in config, but this logic might be useful for other users as well (potential scenarios outlined at the end of comment).

Currently the logic is such that:

- If the environment is not `production`, docs are accessible regardless of `viewApiDocs`
- If the environment is `production` and the user is not authenticated, docs are not accessible regardless of `viewApiDocs`

This PR changes the order so that `viewApiDocs` gate takes priority. Once this gate is defined, docs will be accessible if and only if it allows access. The environment check will only be used as a fallback when the `viewApiDocs` gate is not defined at all. Besides this PR also removes the `(bool) $request->user()` check so that "guest gates" would also be usable.

This will enable scenarios like:

- Adding guest access on prod (for public API docs) e.g. `Gate::define('viewApiDocs', fn($user = null) => true)`
- Disabling and enabling docs via a config switch on any environment, e.g. `Gate::define('viewApiDocs', fn($user) => config('scramble.enabled') && /* other checks */)`
- Adding auth to deployment in a `dev` or `staging` environment, e.g. `Gate::define('viewApiDocs', fn($user) => /* checks */)` (this is currently not checked in such environments)

### DIFF
--- a/docs/usage/access.md
+++ b/docs/usage/access.md
@@ -3,9 +3,9 @@ title: Docs authorization
 weight: 5
 ---
 
-Scramble exposes docs at the `/docs/api` URI. By default, you will only be able to access this route if the environment is not `production`.
+Scramble exposes docs at the `/docs/api` URI. By default, you will only be able to access this route in the `local` environment.
 
-If you need to allow access to docs in `production` environment, implement gate called `viewApiDocs`:
+Define `viewApiDocs` gate if you need to allow the access in other environments:
 
 ```php
 Gate::define('viewApiDocs', function (User $user) {
--- a/src/Http/Middleware/RestrictedDocsAccess.php
+++ b/src/Http/Middleware/RestrictedDocsAccess.php
@@ -8,15 +8,11 @@ class RestrictedDocsAccess
 {
     public function handle($request, \Closure $next)
     {
-        if (array_key_exists('viewApiDocs', Gate::abilities())) {
-            if (Gate::allows('viewApiDocs')) {
-                return $next($request);
-            }
-            
-            abort(403);
+        if (app()->environment('local')) {
+            return $next($request);
         }
-
-        if (! app()->environment('production')) {
+        
+        if (array_key_exists('viewApiDocs', Gate::abilities()) && Gate::allows('viewApiDocs')) {
             return $next($request);
         }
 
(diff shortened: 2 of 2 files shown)


romalytvynenko:
Hey @tontonsb 

Thanks for the suggestion.

When developing this part I thought about the order and I've decided to use whatever familiar I can find in Laravel. So what I found is this: https://laravel.com/docs/9.x/horizon#dashboard-authorization

As you can see, the gate is controlling an access in **non-local** environment.

However, I like the idea about not using `$request->user()` like you've suggested. Also, I'd rather change the logic but keep the priority. I mean, to allow access to docs only in `local` environment and use a gate for non-local ones, so the logic is like in Horizon.

What do you think?

tontonsb:
I agree that following Horizon's example makes sense.

If the `! production` check is changed to `local`, it would still allow all the scenarios that I outlined, so I have no objections to such logic.

If doing the auth like Horizon does, maybe we should also drop the `array_key_exists('viewApiDocs', Gate::abilities())` part? The difference is that `Gate::allows('viewApiDocs')` without checking if the ability exists will allow admins (`Gate::before(/* ... */)`) to use the feature even if the gate is not defined at all. We could even shorten it all to `if (app()->environment('local') || Gate::allows('viewApiDocs'))` :D

romalytvynenko:
@tontonsb Thanks, that looks awesome!

Yeah, it totally makes sense to remove `array_key_exists` check and keep only `allows`.

Speaking about shortening the expression, I'd rather keep them in separate `if`s: https://spatie.be/guidelines/laravel-php#content-compound-ifs

Also, the change to docs needed again, as I've merged that fix about `production` environment from `patch-1` 🙈

Thanks for making these changes and cleaning things up! I really appreciate it!

tontonsb:
OK, dropped `array_key_exists`, let me know if you see any more problems.

> Also, the change to docs needed again, as I've merged that fix about production environment from patch-1 🙈

Yeah, I noticed it as well :) I gave [a try](https://github.com/dedoc/scramble/pull/102/commits/d3cfa14c831ad8c0cced5b6cbd7addfeea3f9e41) to document the behaviour in this PR.

romalytvynenko:
@tontonsb thanks! That's perfect!

PR #103, opened 2023-01-16
Add support for translations in abort messages

dmason30:
This is a POC for parsing ranslations when used in abort helpers:

```php
abort(400, __('response.abort'));
abort(400, __('This is an error'));
abort_if(rand(0, 1) > 0, 400, \Illuminate\Support\Facades\Lang::get('response.abort_if'));
abort_unless(rand(0, 1) > 0, 400, trans('response.abort_unless'));
```

It will check if the string is translatable and if it is use the translated value as an example in the docs.

I have tried to keep it pretty clean, but this is my first time diving into this repo so I appreciate there may have been a better way to approach this.

### DIFF
--- a/src/Support/InferExtensions/AbortHelpersExceptionInfer.php
+++ b/src/Support/InferExtensions/AbortHelpersExceptionInfer.php
@@ -10,8 +10,11 @@ use Dedoc\Scramble\Support\Type\Literal\LiteralStringType;
 use Dedoc\Scramble\Support\Type\ObjectType;
 use Dedoc\Scramble\Support\Type\TypeHelper;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
+use Illuminate\Support\Facades\Lang;
+use PhpParser\Node\Arg;
 use PhpParser\Node\Expr;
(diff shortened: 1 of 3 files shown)


romalytvynenko:
Hey @dmason30 

This looks good to me! I like the idea!

Do you think there are other possible cases of where we may need to use translated strings in documentation?

dmason30:
@romalytvynenko Probably throwing exceptions (not via abort) `throw new \Exception(__(...))`

also throwing via `throw_if` / `throw_unless` etc

romalytvynenko:
I see!

I thought about introducing another type that extends `string literal` type (something like `TranslatedStringLiteral`). This new type will be returned from methods that translate strings.

This way we can stop handling translated strings every time we need to work with them and we only need to handle string literal types.

This can be done via introducing type inferring extension that resolves a type from calling translation functions.

Thoughts?

dmason30:
@romalytvynenko  Sorry for the delay in responding our focus was switched from our new public api for the last couple of months.

Yes probably a specific type class for translations, I only use the translations on a public api for error messages but I assume some will have them in their responses too/

PR #117, opened 2023-03-03
Add caching to docs api route

dmason30:
This is a proof of concept of adding caching to the `docs/api.json` rather than generating the file every time the docs are loaded. 

This can be especially helpful when the docs are deployed to production and won't change until next deployment.

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -27,6 +27,23 @@ return [
         'description' => '',
     ],
 
+    'cache' => [
+        /*
+         * Enable caching of the docs/api.json route
+         */
+        'enabled' => env('SCRAMBLE_CACHE_ENABLED', false),
+
+        /*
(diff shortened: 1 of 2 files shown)


romalytvynenko:
@dmason30 sorry for the late reply.

For now I'm hesitant about adding this as a core part of the library in this way.

I'm postponing caching implementation as I see it enabled by default, based on per-route basis. The cache for a route will be invalidated if any file which affects route's documentation changes. For cache to work like this, some changes still should be made and I'm definitely moving in that direction.

If cache is needed anyway, it can be implemented and added as a middleware into config published in user's project, so that's way there is no changes in the lib.

Please feel free to open another PR to docs with such middleware example (may live in new "Recipes" page in documentation).

Thanks.

PR #140, opened 2023-05-08
Validation rule syntax that supports laravel-actions package

terranc:


romalytvynenko:
Thanks for your suggestion.

Please consider implementing this as an extension and releasing as a separate package as `laravel-actions` is not a Laravel's core part. 

Read more about extensions here: https://scramble.dedoc.co/developers/extensions

tommica:
@terranc +1 would really appreciate if you could turn this into an extensions

tommica:
I made a package for this: https://packagist.org/packages/tommica/laravel-actions-scramble

PR #162, opened 2023-07-29
feat: allow customisation of docs path

bensherred:
This PR adds the ability to customize the route where documentation can be accessed. This is especially helpful when your base API is available at `https://api.example.com`.

Allowing users to change the path would make it possible to access docs from `https://api.example.com/docs` instead of `https://api.example.com/docs/api`.

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -3,6 +3,12 @@
 use Dedoc\Scramble\Http\Middleware\RestrictedDocsAccess;
 
 return [
+    /**
+     * This is the URI path where the docs will be accessible from. Feel free
+     * to change this path to anything you like.
+     */
+    'path' => env('SCRAMBLE_PATH', 'docs/api'),
+
     /*
      * Your API path. By default, all routes starting with this path will be added to the docs.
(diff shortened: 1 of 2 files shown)


romalytvynenko:
Hey @bensherred!

Thanks for the contribution.

There was a feature request, https://github.com/dedoc/scramble/discussions/115, and I was thinking about a bit different API to solve that issue (and this one you're solving).

So I was thinking about introducing a way to manually register scramble docs routes using some new method on `Scramble` class:

```php
// routes/web.php
Route::domain('api.'.config('app.domain'), function () {
    Scramble::docsRoutes(path: 'docs/api');
});
```

This way consumers can pick a domain and also customize path.

Are you open to try it out?

romalytvynenko:
Hey @bensherred 

I was going to explore this sort of API for custom routes:
https://github.com/dedoc/scramble/discussions/174#discussioncomment-6750327

It will allow not only to customize path, but also allow having docs for more API versions.

bensherred:
That makes perfect sense @romalytvynenko

romalytvynenko:
Closing due to inactivity and preference in other approach

PR #189, opened 2023-08-21
Default classes in service provider retrieval via DI container

Wulfheart:
Hello,

I have a use case where I would like to override the default classes and extensions which are registered without having to mess with the bootstrapping of the package itself (read: "I don't want to copy over the `$this->app->when(...` part).

### DIFF
--- a/src/Configuration/DefaultExceptionToResponseExtensionClasses.php
+++ b/src/Configuration/DefaultExceptionToResponseExtensionClasses.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Dedoc\Scramble\Configuration;
+
+use Dedoc\Scramble\Extensions\ExceptionToResponseExtension;
+use Dedoc\Scramble\Support\ExceptionToResponseExtensions\AuthorizationExceptionToResponseExtension;
+use Dedoc\Scramble\Support\ExceptionToResponseExtensions\HttpExceptionToResponseExtension;
+use Dedoc\Scramble\Support\ExceptionToResponseExtensions\NotFoundExceptionToResponseExtension;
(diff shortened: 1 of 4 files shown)


romalytvynenko:
@Wulfheart on the one hand, I like the idea. On the other hand, this really makes these a part of API and will limit me in making more refactoring in the package internals...

Besides support of `query` parameters, why do you need this? What are other cases?

Wulfheart:
None (yet).

Wulfheart:
@romalytvynenko What about making them available via a config key?

romalytvynenko:
@Wulfheart that also feels to me like making it a part of public API which I'd like to avoid.

My vision and expectations for Scramble's extensions system is that you'd literally will be able whatever you need to do, within custom extension.

For example, let's look at that issue you had with parameter placement (in query or in request body). Ideally you should be able to create `OperationExtension`, in which you're able to see the annotation on rule parameter, and move it from request body to query (by changing it's type). If this is not possible via extensions, extensions API should be improved. This is why I love the fact you're sharing the examples and specifics of what you're trying to achieve – I will be able to improve extensions system.

Contrary, if this capabilities are not added to the extensions and people will override core extensions, I will need to treat them as part of public API and will need to think twice before introducing some changes to them, as it may break people's code. I want to avoid this and feel responsible only for making sure that extensions API stay the same, not internals.  

So this is why I'd prefer reviewing case-by-case of what is currently not possible via extensions (or what feels clunky) and improving extensions, rather then allowing to override core extensions. 

Thank you.

If that makes sense to you, please feel free to close the PR.

Wulfheart:
Makes perfect sense for me. Thanks for clarifying.

PR #195, opened 2023-08-23
Added configuration to enable Subresource Integrity (SRI) check

kichetof:
This PR add a configuration to enable Subresource Integrity (SRI) for `stoplight/elements` assets.

As it could be a security issue, I enabled it by default.
Unfortunately, you have to edit the configuration to update UI. _(version and both hash keys)_

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -40,6 +40,21 @@ return [
          * URL to an image that displays as a small square logo next to the title, above the table of contents.
          */
         'logo' => '',
+
+        /*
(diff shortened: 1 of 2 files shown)


romalytvynenko:
Hey @kichetof 

Thank you for the contribution, but I don't plan to merge it at the moment. If this is needed or required on your end, you should be able to publish the docs view and make modifications there.

Thank you!

kichetof:
Thanks @romalytvynenko for your feedback.
I already published the view to add it 😎 

Thanks for this excellent package!

PR #222, opened 2023-09-12
add: Image and file validator support, request @mediaType tag support

WildEgo:
This PR adds detection for 'image' and 'file' rules and setting their format as binary, in addition it also adds support for the @mediaType tag that will set the mediaType of the request closing discusttion #223 .


romalytvynenko:
@WildEgo looks awesome! Thanks you so much for implementation and making some tests. I really appreciate it.

Can you please also add a test for manually setting `@mediaType`? 

I also suggest to rename it to `@responseMediaType` as in OpenAPI spec it is a concept that applies to requests as well.

elmehdiabdi-src:
why you don't merge this ??

PR #236, opened 2023-09-26
Add `whenCounted` support in API resources

adjustmentlayer:
In this PR I added support for the`whenCounted` method in json resources.
```php
public function toArray() {
    return [
        'variantsCount' => $this->whenCounted('variants'),
    ];
}
```


tests/__snapshots__/TypeToSchemaTransformerTest__it_gets_json_resource_type_with_when_counted__1.yml
```diff
@@ -0,0 +1,12 @@
+type: object
+properties:
+    bar_single: { type: integer }
+    bar_fake_count: { type: integer }
+    bar_default: { anyOf: [{ type: integer }, { type: integer }] }
```
romalytvynenko, line 5: @adjustmentlayer I believe this one should be just `{ type: integer }`.

To achieve this, you can update the extension to construct `Union` like so:

```php
            return Union::wrap([
                $this->value(TypeHelper::getArgType($scope, $node->args, ['value', 1])),
                $this->value(TypeHelper::getArgType($scope, $node->args, ['default', 2], new ObjectType(MissingValue::class))),
            ]);
```

This should fix the issue.

romalytvynenko (commented):
Thanks for your contribution!

tests/__snapshots__/TypeToSchemaTransformerTest__it_gets_json_resource_type_with_when_counted__1.yml
adjustmentlayer: Unfortunately, wrapping won't really help in case of different literal types
```php
$this->whenCounted('bar', fn () => 1, 5)
 // Union::wrap([new LiteralIntegerType(1), new LiteralIntegerType(5)]);
 // { anyOf: [{ type: integer }, { type: integer }] }
```
But it will definitely help in case of identical literal types
```php
$this->whenCounted('bar', fn () => 1, 1)
 // Union::wrap([new LiteralIntegerType(1), new LiteralIntegerType(1)]);
 // { type: integer }
```

So I'm thinking, should we reflect the literal value in the schema? 
Like so
`{ anyOf: [{ type: integer(1) }, { type: integer(5) }] }`

romalytvynenko: @adjustmentlayer you're right, that's literals!

Speaking of this: 
> `{ anyOf: [{ type: integer(1) }, { type: integer(5) }] }`

Such annotation is not supported on OpenAPI. 

I had the similar issue with strings, and for them I've added a bit more logic. So when strings literals types are converted to OpenAPI type, the literal value goes into `example`. This way there is one type, but few examples:

`{ type: integer, examples: [1, 5] }`

adjustmentlayer: @romalytvynenko ok, then I need to do the same trick with integer literals that you did with string literals. I will look into that.

adjustmentlayer: @romalytvynenko It is done

romalytvynenko: @adjustmentlayer but it shouldn't be `enum`, it should be `examples`. `enum` literally means that that are only possible values.

adjustmentlayer: That's right. If you have code like this, then 1 and 5 are the only possible values. 
```php
this->whenCounted('bar', fn () => 1, 5)
 // { type: [{ type: integer, enum[1, 5] } ] }
```

adjustmentlayer: I don't really understand what you mean. I just took your logic for string literals and repeated it for integer literals. I thought you wanted me to do this.

adjustmentlayer: But maybe I missed something, I will try looking again

adjustmentlayer: I checked your code again for string literals. 
You also transform a union of string literals to enum, so I don't understand what I did wrong
https://github.com/dedoc/scramble/blob/a7e02cb90eff09a6e751da750d9f72cf256b6d6b/tests/__snapshots__/TypesRecognitionTest__it_handles_unions_of_string_literals_with_(_var_foobarstring_)__1.yml#L3

Here is your commit https://github.com/dedoc/scramble/commit/2250e75f9462d09f1c3d607f19ac3f24bf21e014 (added string literals casting to enum)

romalytvynenko: @adjustmentlayer 

> That's right. If you have code like this, then 1 and 5 are the only possible values.
> $this->whenCounted('bar', fn () => 1, 5)
> // { type: integer, enum[1, 5] }

Thanks, I see. Yeah, that makes sense to me now!

PR #237, opened 2023-09-26
Request validation rules documentation based on static analysis instead of `eval`

### LINKED ISSUE #226: Docs generation fails for requests that utilise custom values
I have the following request validation:
`'email'                 => 'required|email|unique:users,email,' . auth()->user()->id`
What it does is it allows updating the users email address, while ensuring a unique email, but also if the same email(the users email in the table) is being sent then mark it as fine.

The scramble error is because of the auth()-user()-id part:
`Cannot generate request documentation: Attempt to read property "id" on null`

romalytvynenko:
WIP

Fixes #226

### DIFF
--- a/src/Infer/Analyzer/MethodAnalysisResult.php
+++ b/src/Infer/Analyzer/MethodAnalysisResult.php
@@ -10,7 +10,6 @@ class MethodAnalysisResult
     public function __construct(
         public Scope $scope,
         public FunctionLikeDefinition $definition,
-    )
-    {
+    ) {
     }
 }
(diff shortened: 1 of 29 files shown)


benascbr:
@romalytvynenko I would enjoy this fix so much :tada: 

I'm on same boat as https://github.com/dedoc/scramble/issues/226

stefanzweifel:
I know this PR has been open a while, but is there anything that could be contributed from the community to bring this into a release?

stefanzweifel:
@romalytvynenko Nothing specific that results in an error in our current project. Saw the PR and really like the idea of using static analysis.

In our project we were struggling to get the api docs working, as we split our validation rules into instance-specific-request classes; but we could find a solution by using `Request::createFrom()` to make Scramble happy again.

PR #247, opened 2023-10-01
Excluding false positives in route filtering in Generator@getRoutes

Carnicero90:
Str::startsWith($route->uri, config('scramble.api_path', 'api') may actually includes web routes (I.E.: '/apicoltore').
Kinda preventing that, plus making config('scramble.api_path') a little more error forgiving (by trimming leading and trailing slashes before matching it with route URIs).

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -141,7 +141,10 @@ class Generator
                 $routeResolver = Scramble::$routeResolver ?? function (Route $route) {
                     $expectedDomain = config('scramble.api_domain');
 
-                    return Str::startsWith($route->uri, config('scramble.api_path', 'api'))
+                    $routePrefix = trim(config('scramble.api_path', 'api'), '/');
+                    $routeRegex = '/^' . preg_quote($routePrefix, '/') . '((\/.+)|$)/';
+
+                    return preg_match($routeRegex, $route->uri)
                         && (! $expectedDomain || $route->getDomain() === $expectedDomain);
                 };
 
--- a/tests/Generator/Routes/RoutesFilteringTest.php
+++ b/tests/Generator/Routes/RoutesFilteringTest.php
@@ -1,0 +1,97 @@
+<?php
+
+use Dedoc\Scramble\Generator;
+use Dedoc\Scramble\Scramble;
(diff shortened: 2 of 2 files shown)


src/Generator.php
```diff
@@ -141,7 +141,10 @@ private function getRoutes(): Collection
                 $routeResolver = Scramble::$routeResolver ?? function (Route $route) {
                     $expectedDomain = config('scramble.api_domain');
 
-                    return Str::startsWith($route->uri, config('scramble.api_path', 'api'))
+                    $routePrefix = trim(config('scramble.api_path', 'api'), '/');
+                    $routeRegex = '/^' . preg_quote($routePrefix, '/') . '((\/.+)|$)/';
+
+                    return preg_match($routeRegex, $route->uri)
```
romalytvynenko, line 147: Does this regexp equal to this approach? If so, I'd prefer `startsWith`.

```suggestion
                    return Str::startsWith($route->uri, rtrim(config('scramble.api_path', 'api'), '/').'/');
```

tests/Generator/Routes/RoutesFilteringTest.php
romalytvynenko, line 1: Formatting fixes needed (line 12 is redundant empty line, broken formatting on lines 77-79).

Also, using data providers can simplify this test file to just a single test with few data items.

romalytvynenko (commented):
Thanks for the PR. Few questions and suggestions before I merge.

tests/Generator/Routes/RoutesFilteringTest.php
Carnicero90: my bad, forgot to run pint I guess.

src/Generator.php
Carnicero90: I think it should be `Str::startsWith($route->uri.'/', rtrim(config('scramble.api_path', 'api'), '/').'/')`: with that fix yeah, it actually does the very same thing. Gonna change that.

romalytvynenko:
Closing due to inactivity

Carnicero90:
Oh my apologies (and merry Christmas), I completely forgot about that pr: I'll update my branch according to your suggestions and resubmit it

Issue #269, opened 2023-11-03
Error when grouping routes by tags: Undefined array key 0


rolexam:
Hey! 

I'm getting error when trying to group controllers by Tags.

`Error when analyzing route 'GET {someRoute}' (Controller@action): Undefined array key 0`

PR #270, opened 2023-11-03
Fixed error when trying to access tag data

rolexam:
Solves [Issue when grouping by tags.](https://github.com/dedoc/scramble/issues/269)

After debugging, I've found, that 
```php
$classPhpDoc->getTagsByName('@tags')
```
return array 
```
array:1 [▼
  1 => PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode {#5375 ▶}
]
```

So, simple array_values method should solve issue of grouping controllers by tags.

### DIFF
--- a/src/Support/OperationExtensions/RequestEssentialsExtension.php
+++ b/src/Support/OperationExtensions/RequestEssentialsExtension.php
@@ -130,7 +130,7 @@ class RequestEssentialsExtension extends OperationExtension
             return [];
         }
 
-        return explode(',', $tagNodes[0]->value->value);
+        return explode(',', array_values($tagNodes)[0]->value->value);
     }
 
     private function getParametersFromString(?string $str)


romalytvynenko:
@rolexam thanks for the fix! Can you please share the piece of a codebase that produces this error?

rolexam:
Sure!

Here is the controller:
`app/Http/Controllers/SampleController.php`
```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;

/**
 *
 * @tags SampleTag
 *
 */
class SampleController extends Controller
{

    public function action()
    {
        return [
            "test" => "test",
        ];
    }

}
```

Route to controller:
`routes/api.php`
```php
Route::get('api/sample-action', [\App\Http\Controllers\SampleController::class, 'action']);
```

Error when on page `http://localhost:8000/docs/api.json`
```Error when analyzing route 'GET api/api/sample-action' (App\Http\Controllers\SampleController@action): Class "App\Http\Controllers\Controller" not found – {projectLocation}/app/Http/Controllers/SampleController.php on line 12```

**Comment by romalytvynenko** [comment]
@rolexam I'm not sure I'm following, sorry.

This PR fixes the following error: `Error when analyzing route 'GET {someRoute}' (Controller@action): Undefined array key 0`

You created reproduction that produces different error: `Error when analyzing route 'GET api/api/sample-action' (App\Http\Controllers\SampleController@action): Class "App\Http\Controllers\Controller" not found – {projectLocation}/app/Http/Controllers/SampleController.php on line 12`

Can you please share reproduction for the error you claim to fix? Or am I missing something?

Thanks!

**Comment by Montaire** [comment]
Hey @romalytvynenko,
I am facing the same issue as well. Basically none of the `@tags` are working at the moment.

From the reproduction I think he left out the
`use App\Http\Controllers\Controller;`
accidentally.

So it would look like this:

```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;
use App\Http\Controllers\Controller;

/**
 *
 * @tags SampleTag
 *
 */
class SampleController extends Controller
{

    public function action()
    {
        return [
            "test" => "test",
        ];
    }

}
```

Montaire:
Basically the problem is that the array index does not starts from 0 here.

Becouse array_filter keeps the indexes from the previous step!

```php
public function getTags(): array
	{
		return array_filter($this->children, static function (PhpDocChildNode $child): bool {
			return $child instanceof PhpDocTagNode;
		});
	}
```

So I got this as well:
```
array:1 [▼ // vendor\dedoc\scramble\src\Support\OperationExtensions\RequestEssentialsExtension.php:132
  2 => PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocTagNode {#6772 ▼
    +name: "@tags"
    +value: PHPStan\PhpDocParser\Ast\PhpDoc\GenericTagValueNode {#6771 ▶}
    -attributes: []
  }
]
```
So this one fails as only index 2 exists.
`return explode(',', $tagNodes[0]->value->value);`

instead of [0] also a [reset](https://www.php.net/manual/en/function.reset.php)($tagNodes) would work.

PR #277, opened 2023-11-28
Add command to export specification json

princejohnsantillan:
This PR adds a command to export the openapi specification json.

```bash
php artisan scramble:export
```

The idea came from my attempt to use this package on an existing site that's deployed to Vapor. Because of the nature of the documentation page where the specification is built during the request, this caused the request to take longer than 30 seconds for my project. Since Vapor times out after 30 seconds the docs page could not load. With that, it would be great to have the ability to export the specification first and use that file instead. This PR is step 1 to making that possible. 

Solves #276


WildEgo:
This would be pretty useful, actually came up to suggest the same thing, would love to see it merged

WildEgo:
> @princejohnsantillan hey! Thanks for the contribution.
> 
> Quick question, what happens with the file later? Or you export it to `public/api.json` and then simply access it with HTTP call?

Not the author, but I personally would just like to have the file anywhere in the app so I can then use my client generator without having to start up the server in of itself, it's nice and easy for automation, as I've seen that sometimes Scramble doesn't detect the types correctly, filling them as string and I need to generate my client 2 times.

romalytvynenko:
@WildEgo Trying to figure out the best default file location here. Can you please elaborate on how exactly you're going to use the file?

>  it's nice and easy for automation

WildEgo:
> @WildEgo Trying to figure out the best default file location here. Can you please elaborate on how exactly you're going to use the file?
> 
> >  it's nice and easy for automation

I personally have no need for it to be in the public folder, I just install something like `openapi-typescript-codegen` and run it to generate my client, I do all of this in a monorepo, the plan would be to generate the API.json every update and then generate a TS client

princejohnsantillan:
@romalytvynenko  thank you for looking into my PR 😁

1. The file can be used for many other reasons because it is an OpenAPI specification. Like for example importing it in Postman. And I agree with @WildEgo  where it can be "nice and easy for automation", for example if you want to upload a copy of your OpenAPI specs to S3. 
2. I was also thinking of using in on the route file, to act as a cache for the specification. There are use cases like what I have with my Vapor setup where the specification does not need to be generated on every single request. And having it cached like this makes the response time of the docs page faster. 

```
Route::get('docs/api.json', function (Dedoc\Scramble\Generator $generator) {
        $apiFile = config('scramble.export_file');

        if(File::exists($apiFile)){
            return File::json($apiFile);
        }
        
        return $generator();
    })->name('scramble.docs.index');
```

src/Console/Commands/ExportSpecifications.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Execute the console command.
+     */
+    public function handle(Generator $generator): void
+    {
+
+        $specifications = json_encode($generator());
+
+        /** @var string filename */
+        $filename = $this->hasOption('path')
```
logicaldox, line 36: @princejohnsantillan I am trying this and it seems like `$this->hasOption('path')` is always returning `true` which causes the command to fail when a path isn't provided, I would suggest adding a test to cover this as your only test covers passing with the path manually

princejohnsantillan: Thanks @logicaldox. Apparently `hasOption` is not the check I thought it was. It's checking the signature not the input. Your suggestion works perfectly and I applied it. Thanks again!

romalytvynenko (approved):
Looking good!

romalytvynenko:
@princejohnsantillan tests are failing. Will merge after those are green! Thank you!

princejohnsantillan:
> @princejohnsantillan tests are failing. Will merge after those are green! Thank you!

@romalytvynenko tests fixed. Sorry for the delay. I was on holiday break.

Issue #130, opened 2023-04-20
Add ability to customize operation tags


romalytvynenko:
Right now tags are resolved based on class name and tags added to controller's class.

Give an option to customize this behavior.

PR #282, opened 2023-12-05
Improved tag behaviour

WildEgo:
From what I saw tags were not really being parsed, with this PR I removed the old tag behavior and replaced it with the following read tags in the following ways:

Single line docblock (outputs `tags: [MultilineTag,MultilineTag2]`):
```php
    /**
     * @tags MultilineTag,MultilineTag2
     */
```

Multi line docblock (outputs `tags: [MultilineTag,MultilineTag2]`):
```php
    /**
     * @tags MultilineTag
     * @tags MultilineTag2
     */
```

In addition it also adds unique tags key in the main document instead of only the path.


WildEgo:
@romalytvynenko Sorr for the ping, could you review this? I also have an extension I'd like to PR after for headers

romalytvynenko:
@WildEgo sure thing, old behaviors cannot be just removed.

I see Scramble as something that has meaningful defaults.

For now I'm going to close this PR as tags improvements is planned in this issue: https://github.com/dedoc/scramble/issues/130 

As a part of that issue we'll add support for per-method `@tags`, thanks for the idea.

romalytvynenko:
@WildEgo feel free to re-open this PR and make some changes. I can see that that issue applies to your problem as well. 

I wanted to implement method `resolveTagsUsing` to override default behavior. Something like this:

```php
// in service provider
Scramble::resolveTagsUsing(function (RouteInfo $routeInfo, Operation $operation) {
    // ...
});
```

By default, tags resolver will pick controller name as a tag. But using `resolveTagsUsing` users would be able to disable/change the behavior.

Also, feel free to keep the implemented tags resolution for methods - this is a great idea.

WildEgo:
@romalytvynenko I could probably try to do that, I don't have a lot of experience in this service provider stuff, tho it seems a lot for simple things like the stuff I added, it'd also make it so it's easier to follow the OAPI spec since my current setup also adds the tags key in the parent object which from probably my lack of understanding would be harder to setup with the `resolveTagsUsing` method

WildEgo:
> @WildEgo feel free to re-open this PR and make some changes. I can see that that issue applies to your problem as well.
> 
> I wanted to implement method `resolveTagsUsing` to override default behavior. Something like this:
> 
> ```
> // in service provider
> Scramble::resolveTagsUsing(function (RouteInfo $routeInfo, Operation $operation) {
>     // ...
> });
> ```
> 
> By default, tags resolver will pick controller name as a tag. But using `resolveTagsUsing` users would be able to disable/change the behavior.
> 
> Also, feel free to keep the implemented tags resolution for methods - this is a great idea.

Haven't had time to work on this too much, is there any chance we could re-open this and maybe merge? Or is the plan to still using `resolveTagsUsing`?

PR #287, opened 2023-12-21
Extract Format in StringType using @format tag

achyutkneupane:
Hello,

I've made this contribution to add a `@format` tag in `Resource`.

Usage Format:
```php
/**
 * @var string $created_at The created time of the instance @format date-time
 */
'created_at' => $this->created_at,
```

Output:
image

Example response:
image

### DIFF
--- a/src/Support/TypeToSchemaExtensions/FormatExtractorExtension.php
+++ b/src/Support/TypeToSchemaExtensions/FormatExtractorExtension.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Dedoc\Scramble\Support\TypeToSchemaExtensions;
+
+use Dedoc\Scramble\Support\Type\Type;
+use Illuminate\Support\Str;
+use PHPStan\PhpDocParser\Ast\PhpDoc\VarTagValueNode;
+use Dedoc\Scramble\Extensions\TypeToSchemaExtension;
+use Dedoc\Scramble\Support\Generator\Types\StringType;
+
+class FormatExtractorExtension extends TypeToSchemaExtension
+{
+    public function getDescription(Type $type)
+    {
+        $docNode = $type->getAttribute("docNode");
+        if(!$docNode) return false;
+
+        $vars = $docNode->getTagsByName("@var");
+        foreach($vars as $var) {
+            $value = $var->value;
+            if(!($value instanceof VarTagValueNode)) return false;
+
+            return $value->description;
+        }
+    }
+
+    public function shouldHandle(Type $type): bool
+    {
+        $description = $this->getDescription($type);
+        if(!$description) return false;
+        return Str::of($description)->contains('@format');
+    }
+
+    public function toSchema(Type $type): StringType
+    {
+        $description = $this->getDescription($type);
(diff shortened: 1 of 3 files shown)


Dipesh79:
Useful @achyutkneupane

romalytvynenko:
Hey @achyutkneupane 

Thanks, great idea.

Let's make it a separate PHP Doc attribute so it is consistent with how attributes in PHP Doc work:

```
/**
 * @var string $created_at The created time of the instance
 * @format date-time
 */
'created_at' => $this->created_at,
```

And also this needs a test.

Thanks!

achyutkneupane:
> Hey @achyutkneupane
> 
> Thanks, great idea.
> 
> Let's make it a separate PHP Doc attribute so it is consistent with how attributes in PHP Doc work:
> 
> ```
> /**
>  * @var string $created_at The created time of the instance
>  * @format date-time
>  */
> 'created_at' => $this->created_at,
> ```
> 
> And also this needs a test.
> 
> Thanks!

Thank you for the response @romalytvynenko.
I'll create a new attribute for `@format` and update this PR along with the test.

achyutkneupane:
Hey @romalytvynenko,
I've used `@format` as an attribute instead of in `description`.
Also, I tried my best to add a test case with all the formats I could find.
Thank you

src/Support/TypeToSchemaExtensions/FormatExtractorExtension.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $docNode = $type->getAttribute("docNode");
+        $vars = $docNode->getTagsByName("@format");
+        $format = reset($vars)->value->value;
+        if(Str::contains($format, "|")) {
+            $format = explode("|", $format);
+        }
+        if(Str::contains($format, ",")) {
+            $format = explode(",", $format)[0];
+        }
```
romalytvynenko, line 45: Seems like a bug can happen here: in a case when format has `|` char, on line 41 it will become an array. Then on line 42 you still keep treating it like a string, which I afraid can result in a bug.

tests/ResourceFormatTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+    return $withFormat->value->value;
+}
+
+it('handles string type with format', function ($phpDoc, $expectedFormat) {
+    $result = getFormatFromDoc($phpDoc);
+
+    expect($result)->toBe($expectedFormat);
+})->with([
+    ["/** @var string \n * @format email */", "email"],
```
romalytvynenko, line 34: After closer review of this test, the test is not really testing newly added extension's code.

romalytvynenko:
@achyutkneupane thanks for the PR, but for now I'm going to close it due to possible issues with implementation and missing test for extension (the existing test is not testing newly added code). This feature also fits well into something I'm working on: support of `@example` tag. 

I really appreciate you taking time and creating the extension, but as a maintainer I want to make sure the newly added features match nicely with an existing work in progress.

I believe this extension is cool and I'm thinking to make some sort of a directory of community-submitted extensions.

Thanks for the contribution!

PR #290, opened 2024-01-02
Add ability to customize doc url using config

ashiquzzaman33:
This PR will give users the ability to change the default doc url.

For example, from http://example.com/docs/api to http://example.com/new/api.

**Why:**

Sometimes, the url conflicts with other packages' default urls, such as [Enlighten](https://github.com/StydeNet/enlighten).

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -15,6 +15,11 @@ return [
      */
     'api_domain' => null,
 
+    /**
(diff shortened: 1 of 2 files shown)


romalytvynenko:
Closing this PR in favor of future change that will allow multiple docs versions (hence paths).

PR #292, opened 2024-01-05
Add deprecation ability

sc-bruno:
This change extends the ability of the `Generator` to accept the deprecation property in the `toArray` method.

I have extended the `Operation` class, I needed this in my custom extension that adds the ability to display a deprecation notice if the method or the class is deprecated with a `@deprecated` comment.

### DIFF
--- a/src/Support/Generator/Operation.php
+++ b/src/Support/Generator/Operation.php
@@ -16,6 +16,8 @@ class Operation
 
     public string $summary = '';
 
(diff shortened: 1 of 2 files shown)


sc-bruno:
This is what it looks like
image

romalytvynenko:
@sc-bruno This is awesome, thanks! Can you please add a simple test?

PR #295, opened 2024-01-08
Fix Param Enums Int

oktayyilmaz:
Fix Param Enums Int


romalytvynenko:
Sorry, cannot accept this pr due to unneeded changes: composer.json change, formatting changes). Also not getting what is the issue this PR fixes.

PR #296, opened 2024-01-09
Add @deprecated tag support

sc-bruno:
This PR adds deprecation support. It accepts the `@deprecated` tag on the class and the method level.
If the class is deprecated, all the class methods are deprecated automatically. 
This can be overridden with the `@not-deprecated` tag. In that case, that method will not get a deprecation notice.

If there is a description for the tag it will be picked up by the extension. In the example, the class and the method have the example description.

```
/**
 * @deprecated Class deprecation description
 */
class FooController extends Controller
{
    /**
     * @deprecated Method deprecation description
     */
    public function store() {}
}
```

## **Example**
image


finalgamer:
@romalytvynenko Hello I have noticed that the tests failed due to an unrelated error

`Error: Internal server error occurred while resolving "actions/checkout@v3". Internal server error occurred while resolving "shivammathur/setup-php@v2"`

could you retry all the jobs?

PR #313, opened 2024-02-20
Manage phone format for examples

Takachi67:
Add a custom example rule using "tel:+337xxxxxxxx" for the format


romalytvynenko:
Sorry, not following what this is for.

PR #314, opened 2024-02-22
Support DBAL-unsupported Database Drivers. 

NiekNijland:
As mentioned [here](https://github.com/dedoc/scramble/discussions/252) this package does not support database drivers that are not supported by Doctrine DBAL. 

The sole problem for this lies within the Model attributes functionality. In order to quickly circumvent this, this PR adds a config option that can disable this functionality. 

This way all the core functionality of this pretty package can be used in combination with for example [MongoDB](https://github.com/mongodb/laravel-mongodb).

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -74,5 +74,7 @@ return [
         RestrictedDocsAccess::class,
     ],
 
+    'disable_model_attributes' => false,
+
     'extensions' => [],
 ];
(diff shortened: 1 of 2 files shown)


romalytvynenko:
Closing this PR in favor of more automatic fix: when there is that DBAL exception, we can automatically prevent it from bubbling up and fallback to `collect()`. This way there won't be any additional configuration.

benascbr:
> in favor of more automatic fix
@romalytvynenko @NiekNijland by any chance do you know which version has this automatic fix
I'm so keen to try this package :tada:

NiekNijland:
@benascbr Checking the change log I don't think it was implemented before L9 support was dropped. My bet it is was fixed in [v0.10.4](https://github.com/dedoc/scramble/releases/tag/v0.10.4)

Issue #329, opened 2024-03-11
Unauthenticated routes have incorrectly generated security setup


manuel-watchenterprise:
If I add `@unauthenticated` tag to the docs of the router action the generated specs is incorrect. The empty security is generated as `[]` instead of `{}`.

```
...
      "422": {
        "$ref": "#/components/responses/ValidationException"
      }
    },
    "security": [
      []
    ]
  }
},
"/auth/register": {
  "post": {
...
```

PR #330, opened 2024-03-11
Fixed empty security descriptor

manuel-watchenterprise:
Empty security descriptor must be an object (`{}`) instead of empty empty array (`[]`).

Closes: #329


romalytvynenko:
@manuel-watchenterprise thank you!

PR #336, opened 2024-03-13
Make TryIt work with Sanctum's cookie based authentication out of the box

### LINKED ISSUE #286: auth:sanctum is not detected or documented?
I have a Laravel 9 project (php 8). Most of our APIs (api/*) are using `auth:sanctum` in the `routes/api.php`.  but the api documents generated by the package don't seem to reflect it. I followed the installation instructions. Is there anything missing?

Meanwhile,  I get a 419 error: CSRF token mismatch when trying 'Send API request' for POST calls. any solution other than commenting off `EnsureFrontendRequestsAreStateful::class,` in Kernel.php? I hope to use it on the production so the API documents can be online for others (access controlled by the gate, of course). Thanks.

vovarudomanenko:
Fixes #286 

In this PR I added code that intercepts every TryIt request and adds an `XSRF-TOKEN` header, which is necessary for Sanctum's cookie based authentication to work correctly. Intercepting was done by monkey-patching `fetch` function

### DIFF
--- a/resources/views/docs.blade.php
+++ b/resources/views/docs.blade.php
@@ -6,8 +6,8 @@
     <title>{{ config('app.name') }} - API Docs</title>
 
     <script src="https://unpkg.com/@stoplight/elements/web-components.min.js"></script>
-
     <link rel="stylesheet" href="https://unpkg.com/@stoplight/elements/styles.min.css">
+
     <script>
         const originalFetch = window.fetch;
 
@@ -37,7 +37,7 @@
             const csrfToken = getCookieValue(CSRF_TOKEN_COOKIE_KEY);
             if (csrfToken) {
                 const { headers = new Headers() } = options || {};
-                updateFetchHeaders(headers, CSRF_TOKEN_HEADER_KEY, unescape(csrfToken));
+                updateFetchHeaders(headers, CSRF_TOKEN_HEADER_KEY, decodeURI(csrfToken));
                 return originalFetch(url, {
                     ...options,
                     headers,
--- a/resources/views/docs.blade.php
+++ b/resources/views/docs.blade.php
@@ -6,7 +6,47 @@
     <title>{{ config('app.name') }} - API Docs</title>
 
     <script src="https://unpkg.com/@stoplight/elements/web-components.min.js"></script>
+
     <link rel="stylesheet" href="https://unpkg.com/@stoplight/elements/styles.min.css">
+    <script>
+        const originalFetch = window.fetch;
+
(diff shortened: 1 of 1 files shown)


vovarudomanenko:
Previously I've been fixing this issue but I've only fixed one part of it. This PR fixes second and the last part. Now Sanctum cookie based authentication should work out of the box.

romalytvynenko:
@vovarudomanenko you are the best, thanks! Will test for API token based auth soon and merge. Thanks again, this was really painful issue.

vovarudomanenko:
@romalytvynenko First of all, you are the best. Secondly, I wanted to tell that you should only test for cookie-based authentication because my solution won't work for token-based one. 

Token-based auth requires a different solution.
Specifically, it requires to pass a token in the `Authorization` header as a bearer token.

I currently don't know how to solve this problem universally for everyone. Because in order to automatically pass a token to a request I must to know how to create the token and that depends on the specific API implementation.

But I know one solution to this problem, which I don't really like, but it works. This solution requires adding the `Authorization` header to the OpenAPI definition and manually inserting the token into TryIt form before submitting a request. 
```php
Scramble::extendOpenApi(function (OpenApi $openApi) {
    $openApi->secure(SecurityScheme::http('bearer', 'JWT'));
});
```
image

romalytvynenko:
@vovarudomanenko hey!

I meant in general if the solution affects token based auth (due to `fetch` fn being patched). I've just tested the behavior and it looks good! 

Speaking of inserting API token automatically – this is really a nice idea. If it'll resurface, I'll add it.

Thanks again for your time. Now Sactum's auth works like a charm!

AucT:
I have group in group

The first group is for prefix
and inner group is
`Route::group(['middleware' => ['auth:sanctum', 'verified', 'email']],`
Scramble doesn't detect that this routes require sanctum token

erkinboy-botirov:
In **resources/views/docs.blade.php** at line 40:

Shouldn't **decodeURIComponent** be used instead of **decodeURI**?

It seems like the XSRF-TOKEN that comes as a cookie is base64 encoded. From [what I found](https://stackoverflow.com/questions/61439540/strange-url-encoding-characters-at-the-end-of-csrf-tokens-laravel-sanctum), an equals sign is added as padding at the end of a base64 encoded string to match a specific number of characters. When this token is passed as a cookie value, it is URL encoded, which turns the equals signs into `%3D`.

As you can see in the picture, the `decodeURI` function won't decode it back, but `decodeURIComponent` will.

Screenshot 2024-05-28 at 10 14 40

PR #339, opened 2024-03-14
Fix `confirmed` rule not working properly when it is used on multiple attributes

nicolaubrasil:
The [PR](https://github.com/dedoc/scramble/pull/124) added support for the confirmed rule, but when there are multiple confirmed rules, it only identifies one of the fields.

This PR aims to provide support for more than one confirmed rule.


romalytvynenko:
Hey @nicolaubrasil! Thanks for this improvement and huge thanks for tests specifically. Good stuff!

PR #341, opened 2024-03-15
Resolve ternary types properly

kburton-dev:
I noticed that some ternary types in my JSON resources were not being resolved properly. For example, a resource could look like this

```
public function toArray(): array
{
    return [
        'foo' => $this->foo ? FooResource::make($this->foo) : null
    ];
}
```

Would get a property type such as `foo: { type: string }`, losing a lot of information about the nested resource.

I hacked together a quick fix for this. I hope I did so in a suitable fashion.

### DIFF
--- a/src/Infer/Scope/Scope.php
+++ b/src/Infer/Scope/Scope.php
@@ -19,6 +19,7 @@ use Dedoc\Scramble\Support\Type\Reference\PropertyFetchReferenceType;
 use Dedoc\Scramble\Support\Type\SelfType;
 use Dedoc\Scramble\Support\Type\TemplateType;
 use Dedoc\Scramble\Support\Type\Type;
+use Dedoc\Scramble\Support\Type\Union;
 use Dedoc\Scramble\Support\Type\UnknownType;
 use PhpParser\Node;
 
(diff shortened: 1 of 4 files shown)


WildEgo:
Could you check how it generates ``$this->when``? I'm not sure how dedoc handles this

kburton-dev:
> Could you check how it generates ``$this->when``? I'm not sure how scramble handles this

It works fine using `->when()` ✅

But then I need to pass a function `fn () => null` as the third argument. With only `null` (the default) the property is omitted in the response.

Using a ternary is pretty common, so I think it should be handled well.

src/Infer/Scope/Scope.php
```diff
@@ -52,6 +53,13 @@ public function getType(Node $node): Type
             return (new ConstFetchTypeGetter)($node);
         }
 
+        if ($node instanceof Node\Expr\Ternary) {
+            return new Union([
+                $this->getType($node->if),
+                $this->getType($node->else)
+            ]);
```
romalytvynenko, line 60: There is `Union::wrap` method that will generate `Union` type in a smart way (it'll remove duplicated types etc):

```suggestion
            return Union::wrap([
                $this->getType($node->if),
                $this->getType($node->else)
            ]);
```

tests/TypeToSchemaTransformerTest.php
```diff
@@ -133,6 +133,15 @@
     ]);
 });
 
+it('gets json resource type with ternary value type', function () {
```
romalytvynenko, line 136: Nice! Thanks for the test!

romalytvynenko (changes requested):
Looking good! Small fix is needed to avoid duplicated sub-types in resulting `Union` and I will happily merge it.

Thanks.

PR #345, opened 2024-03-18
Determine eloquent attribute type from casts

kburton-dev:
As mentioned in [this discussion](https://github.com/dedoc/scramble/discussions/278), the schema types differ slightly when generating the OpenAPI json, if using different databases, such as MySQL and SQLite.

This is to be expected if the idea is to rely on the DB schema alone to generate the API schema from Eloquent models.

This PR aims to add support for the various model casts that Laravel provides.


romalytvynenko:
@kburton-dev what a great work!

Thank you so much for your time, I appreciate it!

PR #346, opened 2024-03-19
Remove request body from `HEAD` and `DELETE`  by default

manuel-watchenterprise:
For specific HTTP methods. For example, DELETE is not officially supported to have request body, however, only GET methods are prevented to have empty request bodies in the output.

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -80,4 +80,17 @@ return [
     ],
 
     'extensions' => [],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Disallow Request Body
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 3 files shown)


manuel-watchenterprise:
@romalytvynenko Actually, I would rather make the default restrict to be `['get', 'head', 'delete']` and so anyone will be able to overwrite this with e.g. `['get']` if they will. But, current implementation is the way around to be completely backward compatible.

romalytvynenko:
Hey @manuel-watchenterprise 

Thanks for the PR. For now I decided to keep it as simple as possible while technically changing Scramble behavior – `delete` and `head` will not have request bodies. Testing with real API shows that Laravel simply won't "see" params for these methods if sent in request body.

PR #347, opened 2024-03-19
Give preference to user-defined extensions over built-in extensions

korridor:
**This PR solves the following problem:**

I tried to create an extension for specific `ResourceCollection`s that implement an interface. This was not possible without these changes since the built-in extensions are executed after the custom ones and the built-in `JsonResourceTypeToSchema` extension would overrule my custom extension.

### DIFF
--- a/src/ScrambleServiceProvider.php
+++ b/src/ScrambleServiceProvider.php
@@ -133,7 +133,7 @@ class ScrambleServiceProvider extends PackageServiceProvider
             return new TypeTransformer(
                 $this->app->make(Infer::class),
                 new Components,
-                array_merge($typesToSchemaExtensions, [
+                array_merge([
                     EnumToSchema::class,
                     JsonResourceTypeToSchema::class,
                     ModelToSchema::class,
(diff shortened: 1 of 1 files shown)


fgiova:
@romalytvynenko no interest in this small feature? The fixed '/docs/' route is very restrictive.

korridor:
@fgiova I don't really understand your comment. How is my PR related to the `/docs/` endpoint?

fgiova:
> @fgiova I don't really understand your comment. How is my PR related to the `/docs/` route?

Sorry my fault

korridor:
@fgiova No worries :)

romalytvynenko:
@korridor I believe currently Scramble gives preference to user's extension. Whatever extension that handled the type first is preferred. Hence if user's extension handled the type first, it will be preferred. 

With your change this behavior will be lost.

Can you describe your case with examples of your code and what you've tried to achieve? I'm closing this PR but let's continue discussion here. I will reopen if I misunderstood.

korridor:
@romalytvynenko Sorry for the delayed response.

In my tests, the opposite is the case. For me, my extension is overwritten by the system extension. Only if I apply this change will my extension work.

I'm actually using your extension in an open-source project called [solidtime](https://github.com/solidtime-io/solidtime/), so I can show you what I'm doing:

 - [PaginatedResourceCollectionTypeToSchema](https://github.com/solidtime-io/solidtime/blob/main/app/Extensions/Scramble/PaginatedResourceCollectionTypeToSchema.php)
 - [ApiExceptionTypeToSchema](https://github.com/solidtime-io/solidtime/blob/main/app/Extensions/Scramble/ApiExceptionTypeToSchema.php)

korridor:
@romalytvynenko Did you have a chance to look at this issue again?

PR #356, opened 2024-04-07
Request params inference from method calls on request, default request values support

### LINKED ISSUE #340: Add ability to set a default value in docs for params
### Discussed in https://github.com/dedoc/scramble/discussions/266

<div type='discussions-op-text'>

<sup>Originally posted by **slnw** October 25, 2023</sup>
Hi, I have response param $foo which default value is 20. How do I set it in the openapi docs?</div>

romalytvynenko:
fixes #340


jaimyborgman:
@romalytvynenko can we extend this with having /** @default */ working for ApiResourceItems?

Like for instance:

```
/**
 * @property Product $resource
 */
class ProductResource extends ResourceDto
{
    public function toArray(Request $request): array
    {
        $resource = $this->resource;

        return [
            /** @format uuid */
            'id' => $resource->getId(),
            ...
            /** @default $status draft */
            'status' => $resource->status,
        ];
    }
}

```

jaimyborgman:
@default would suit this case better since i'd like to define and let know that there is a default value for "status" instead of giving an example :)

PR #374, opened 2024-05-17
Ability to set favicon for the docs website

valpuia:
This PR introduce 'favicon' in config file so the users can set the documentations's website favicon

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -20,21 +20,6 @@ return [
      */
     'export_path' => 'api.json',
 
-    /**
-     * Favicon of the documentations's website.
-     */
-    'favicon' => [
-        /**
-         * The URL of the favicon image
-         */
-        'href' => '',
-
-        /**
-         * Type attribute to specify the favicon image type
-         */
-        'type' => '',
-    ],
-
     'info' => [
         /*
          * API version.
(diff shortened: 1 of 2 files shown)


romalytvynenko:
To be honest, I think that the best way to do this for now is to publish a view by running php artisan vendor:publish --provider="Dedoc\Scramble\ScrambleServiceProvider" --tag="scramble-views" and make changes to the view.

IMO, while introducing additional config seems tempting, there are a lot of things to favicon! There are favicon different sizes, there is apple touch icon, etc. So I'd rather publish a view.

Let me know if that makes sense.

valpuia:
I don't have any apple product so I am not aware of any apple related :laughing:, but normally the image (favicon) itself is made for exact size (at least for me) but yeah publishing a view and include manually will make control easily

valpuia:
> I don't have any apple product so I am not aware of any apple related 😆, but normally the image (favicon) itself is made for exact size (at least for me) but yeah publishing a view and include manually will make control easily

Sorry, was thinking different. I thought if I use `16*16` then it'll work on most of them..

valpuia:
@romalytvynenko what do you think like this, will make sense and covers most of them:

``` php
//config
'favicons' => [
    '16' => [
        'href' => '',
        'rel' => '',
        'type' => '',
    ],

    '32' => [
        'href' => '',
        'rel' => '',
        'type' => '',
    ],
],
```
and in blade
```html
@foreach ($config->get('favicons') as $key => $item)
    <link rel="{{ $item['rel'] }}" sizes="{{ $key.'x'.$key }}" type="{{ $item['type'] }}" href="{{ $item['href'] }}">
@endforeach
```

romalytvynenko:
@valpuia yeah, sure this is possible. But it complicates already complex config. Also, won't it be simpler to publish a view in this case? IMO it would be

PR #375, opened 2024-05-17
Improve inferring model property types, based on DB driver

kburton-dev:
This is an attempt to more accurately infer the PHP type that is mapped by the underlying DB connection.

I did not test the types returned when using SQL Server connection, since I have never seen a Laravel app use this driver in the wild.

### DIFF
--- a/src/Support/InferExtensions/ModelExtension.php
+++ b/src/Support/InferExtensions/ModelExtension.php
@@ -56,7 +56,7 @@ class ModelExtension implements MethodReturnTypeExtension, PropertyTypeExtension
 
         if ($attribute = $info->get('attributes')->get($event->getName())) {
             $baseType = $this->getAttributeTypeFromEloquentCasts($attribute['cast'] ?? '')
-                ?? $this->getAttributeTypeFromDbColumnType($attribute['type'] ?? '');
+                ?? $this->getAttributeTypeFromDbColumnType($attribute['type'], $attribute['driver']);
 
             if ($attribute['nullable']) {
                 return Union::wrap([$baseType, new NullType()]);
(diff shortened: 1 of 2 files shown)


src/Support/InferExtensions/ModelExtension.php
```diff
@@ -72,22 +72,34 @@ public function getPropertyType(PropertyFetchEvent $event): ?Type
...
+        if ($columnType === null) {
+            return new UnknownType();
+        }
+
+        $typeName = str($columnType)
+            ->before(' ') // strip modifiers from a type name such as `bigint unsigned`
+            ->before('(') // strip the length from a type name such as `tinyint(4)`
+            ->toString();
+
+        if (in_array($typeName, ['int', 'integer', 'tinyint', 'smallint', 'mediumint', 'bigint'])) {
```
kburton-dev, line 90: Wondering if I need to specify `smallserial, serial, bigserial` for `pgsql`

src/Support/InferExtensions/ModelExtension.php
```diff
@@ -72,22 +72,34 @@ public function getPropertyType(PropertyFetchEvent $event): ?Type
...
-        $attributeType = match ($typeName) {
-            'int', 'integer', 'bigint' => new IntegerType(),
-            'float', 'double', 'decimal' => new FloatType(),
-            'varchar', 'string', 'text', 'datetime' => new StringType(), // string, text - needed?
-            'tinyint', 'bool', 'boolean' => new BooleanType(), // bool, boolean - needed?
-            'json', 'array' => new ArrayType(),
-            default => new UnknownType("unimplemented DB column type [$type]"),
-        };
+        if ($columnType === null) {
+            return new UnknownType();
```
romalytvynenko, line 82: Should we keep info about a column type being not implemented? Current implementation returns `StringType` if type is not handled which may hide debugging info in such cases.

src/Support/InferExtensions/ModelExtension.php
romalytvynenko: @kburton-dev If you know how these types are mapped to PHP types, go ahead :-)

romalytvynenko (commented):
Great work as always. One question about keeping debug info in `UnknownType`

src/Support/InferExtensions/ModelExtension.php
kburton-dev: Actually, I tested this now, and since those `serial` types are not true types (just notational convenience in pgsql), they are  just `integers` under the hood

PR #391, opened 2024-05-22
`match` node type inference

WildEgo:
This is a basic implementation of `match` parsing, I have no experience with testing stuff like this if there's a way to handle this better please do let me know.

### DIFF
--- a/src/Infer/Scope/Scope.php
+++ b/src/Infer/Scope/Scope.php
@@ -61,14 +61,11 @@ class Scope
         }
 
         if ($node instanceof Node\Expr\Match_) {
-            return match (get_class($node->cond)) {
-                Node\Expr\ArrayDimFetch::class, Node\Expr\Variable::class => Union::wrap(
-                    collect($node->arms)
-                        ->map(fn (Node\MatchArm $arm) => $this->getType($arm->body))
-                        ->toArray()
-                ),
-                default => new UnknownType('Cannot infer type of property fetch: not supported yet.'),
-            };
+            return Union::wrap(
+                collect($node->arms)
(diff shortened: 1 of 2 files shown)


src/Infer/Scope/Scope.php
```diff
@@ -60,6 +60,17 @@ public function getType(Node $node): Type
...
 
+        if ($node instanceof Node\Expr\Match_) {
+            return match (get_class($node->cond)) {
+                Node\Expr\ArrayDimFetch::class, Node\Expr\Variable::class => Union::wrap(
+                    collect($node->arms)
+                        ->map(fn (Node\MatchArm $arm) => $this->getType($arm->body))
+                        ->toArray()
+                ),
+                default => new UnknownType('Cannot infer type of property fetch: not supported yet.'),
+            };
```
romalytvynenko, line 71: Do we really need to do `match` condition type check? Won't simply wrapping all arms types into `Union` be sufficient?
```suggestion
            return Union::wrap(
                collect($node->arms)
                    ->map(fn (Node\MatchArm $arm) => $this->getType($arm->body))
                    ->toArray()
            );
```

WildEgo: I do think this might make sense, I'm mostly just afraid there's a return of a type that is not any of those two, but if we know that that'll never be the case that's perfect

romalytvynenko: @WildEgo can you show some examples of PHP code you you've this PR for?

romalytvynenko: @WildEgo gotcha. You definitely can keep just `Union` and skip checking for condition field.

PR #399, opened 2024-05-26
Fix form requests instantiation when request is type hinted using bound interface

kennyhorna:
First of all, great package! 

#### Problem I faced
When I was implementing the repo, I noticed that some of the request details were not generated because the request object was being type-hinted using an interface. This can happen within packages when the request interface is bounded to the container through a service provider, allowing the user to then override the implementation through a config key. 

This was what the docs displayed:

screenshot_2024-05-26 at 13 57 50

#### Solution
Playing with the source code I added a way to, in case it detects that it's not a class name but an interface one, it pulls it from the container (if mapped).

Now it handles this edge case gracefully:

screenshot_2024-05-26 at 14 00 28

### DIFF
--- a/src/Support/OperationExtensions/RulesExtractor/FormRequestRulesExtractor.php
+++ b/src/Support/OperationExtensions/RulesExtractor/FormRequestRulesExtractor.php
@@ -10,6 +10,7 @@ use PhpParser\Node;
 use PhpParser\Node\FunctionLike;
 use PhpParser\Node\Param;
 use PhpParser\NodeFinder;
+use ReflectionClass;
 
 class FormRequestRulesExtractor
 {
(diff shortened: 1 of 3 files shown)


romalytvynenko:
@kennyhorna thank you!

I'm wondering, won't this issue be solved if you simply change the form request instance creation logic? Currently the instance is created simply using `new` keyword:
https://github.com/dedoc/scramble/blob/19d8da29f8a5df34c14d9d50f66c10f47956d066/src/Support/OperationExtensions/RulesExtractor/FormRequestRulesExtractor.php#L55

I feel like changing this line so the instance is created by service container may solve the problem as well as feels simpler:
```php
$request = app($requestClassName);
```

kennyhorna:
@romalytvynenko I thought about that and tried it, but doing it in that way it evaluates the request rendering it like this:

screenshot_2024-05-28 at 09 51 15

I also tried this with the same result:

```php
$request = app()->make($requestClassName);
```

That's why I ended up with the proposed solution

PR #424, opened 2024-06-19
support parameters serialization

layerok:


layerok:
Actually, it turned out that I didn't need this functionality. I initially thought that specifing `style="form"` and `explode=false` would change presentation of query parameters in spotlight elements docs. But that didn't happened. Nothing changed. So in my case I don't need this functionality anymore. However I don't think the ability to configure `style` and `explode` attributes will be excessive. Regardless of whether spotlight is able to present query parameters differently depending on `style` and `explode` attributes or not, scrumble will be able to generate more complete openapi specification.

PR #471, opened 2024-07-29
Allow operation extensions

shokme:
Some tools like Spotlight can accept custom attributes (e.g: x-internal).

### DIFF
--- a/src/Support/Generator/Operation.php
+++ b/src/Support/Generator/Operation.php
@@ -5,6 +5,7 @@ namespace Dedoc\Scramble\Support\Generator;
 class Operation
 {
     use WithAttributes;
+    use WithExtensions;
 
     public string $method;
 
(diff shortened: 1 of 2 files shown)


shokme:
Ok, copied from Parameter 👍🏻

shokme:
> @shokme Awesome! Can you share how you plan to use it?

Sure, this `extension` will allow stoplight to protect non-public routes
```php
class ScrambleInternalRoutesExtension extends OperationExtension
{
    public function handle(Operation $operation, RouteInfo $routeInfo)
    {
        if (count($routeInfo->phpDoc()->getTagsByName('@public'))) {
            $operation->setExtensionProperty('internal', false);
        } else {
            $operation->setExtensionProperty('internal', true);
        }
    }
}
```

PR #498, opened 2024-08-23
fix: issue with extensions being overwritten by built-in extensions

horlarme:
This is a fix for this issue/question https://github.com/dedoc/scramble/discussions/442

The order of the types of extensions caused the issue. All the other extension types are merged by putting the custom ones at the end of the array. Still, in the case of `TypeTransformer` it was at the beginning which caused custom extensions to be overridden by the built-in extensions.

### DIFF
--- a/src/ScrambleServiceProvider.php
+++ b/src/ScrambleServiceProvider.php
@@ -172,7 +172,7 @@ class ScrambleServiceProvider extends PackageServiceProvider
             return new TypeTransformer(
                 app()->make(Infer::class),
                 new Components,
-                array_merge($typesToSchemaExtensions, [
+                array_merge([
                     EnumToSchema::class,
                     JsonResourceTypeToSchema::class,
                     ModelToSchema::class,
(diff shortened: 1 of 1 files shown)


hn-seoai:
I think the _real_ issue we're having, is that we can't choose exactly when our extensions are being used. I'm thinking a priority property, a before/after hook, or some resolution logic to replace one of the default extensions would be ideal. In my case, I still want the `EnumToSchema`, `JsonResourceTypeToSchema`, `ModelToSchema`, and `EloquentCollectionToSchema` to be called before my `PaginatedResourceCollectionTypeToSchema`.

The simplest solution would be for the package classes to have `public int $priority = 10; // increments of 10`, allowing us to put 9 (or more if same priority won't break functionality) extensions in between each of the defaults. Custom extensions would have a default priority of 1, so the existing behavior is unchanged.

If no one else implements this, I'll start working on it this weekend.

@romalytvynenko Would this be an acceptable solution?

PR #509, opened 2024-08-30
Fixed: use container to make requests for extraction

james2doyle:
FormRequest that have a binding in container will not be properly instantiated when the docs are being generated.

Screen Shot 2024-08-30 at 11 18 16 AM

The solution proposed in this comment seems to work for me:

https://github.com/dedoc/scramble/pull/399#issuecomment-2134314719

Resolving the class from the container instead of _newing_ it up, works.

I don't see the other error where "it evaluates the request rendering".

The change to using `app` breaks the tests though. It tries to actually create the FormRequest which causes the validation to trigger. At least that is what it seems like.

Any thoughts on how to proceed here? I can take a crack at a PR, if that helps. But I'm not sure what to fix....

### DIFF
--- a/src/Support/OperationExtensions/RulesExtractor/FormRequestRulesExtractor.php
+++ b/src/Support/OperationExtensions/RulesExtractor/FormRequestRulesExtractor.php
@@ -71,7 +71,7 @@ class FormRequestRulesExtractor
         $requestClassName = $this->getFormRequestClassName();
 
         /** @var Request $request */
-        $request = (new $requestClassName);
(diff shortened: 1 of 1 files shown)


james2doyle:
@romalytvynenko Yeah, I understand. I tried to write some code that would step through the container binding and try to find the classes that are injected. I'm not very experienced working with reflection though, so I couldn't figure it out. There were some examples of people getting it working but those were other classes that didn't trigger functionality when they were resolved.

PR #512, opened 2024-08-31
Added: support for redirects

james2doyle:
Not sure if this is the correct way to add something like this.

Let me know if there is anything to change.


james2doyle:
Ok. So is this PR stale then?

PR #517, opened 2024-09-04
Fixed incorrect data wrap key when using `AnonymousResourceCollection` and redefining it in collected resource

axelrindle:
Accesses the `$wrap` property on the `$collectingResourceType` directly to decide whether to wrap a resource collection in a `data` array or not.


axelrindle:
Certainly. I have a resource where I set the static `$wrap` property to `null`. This results in the following JSON response (example):

```json
[
  {
    "type": "attachment",
    "url": "..."
  }
]
```

Scramble however incorrectly displays the following structure:

```json
{
  "data": [
    {
      "type": "attachment",
      "url": "..."
    }
  ]
}
```

I fixed it by directly accessing the `$wrap` property on the collection type (`$collectingResourceType`) in the class `AnonymousResourceCollectionTypeToSchema`. Otherwise `AnonymousResourceCollection::$wrap` will always return `'data'` instead of what's defined by the actual resource.

PR #582, opened 2024-10-15
Change default OpenAPI JSON export to pretty print

coffee-r:
see https://github.com/dedoc/scramble/discussions/581


hn-seoai:
I'm not entirely sure why pretty print should be the default. Wouldn't it make more sense to have a `--pretty` option?

I rarely look at the json spec, it's simply being passed to whatever codegen tool the project uses. More than doubling the filesize with whitespaces seems unnecessary.

PR #594, opened 2024-10-22
Ability to exclude routes using attributes instead of re-defining routes resolution method

romalytvynenko:


makroxyz:
Amazing job!!!!

PR #674, opened 2025-01-07
fix: Symfony\Component\HttpKernel\Exception\ types

phpsa:
using this example docblock:

```
    /**
     * Flags the policy for automatic renewal
     *
     * @param string $policyId the policy id to flag for renewal
     *
     * @return array{status:bool}
     * @response array{status:true}
     *
     * @throws BadRequestHttpException|NotAcceptableHttpException
     */
```

to have automated responses mapped over:
image
image


phpsa:
@romalytvynenko - thanks for the feedback, 
have removed NotFound and added a test to confirm they are pickedup.

PR #677, opened 2025-01-09
feat: support system ui theme

YournameITManoah:


YournameITManoah:
@romalytvynenko could you take a look at this small but very nice feature?

PR #680, opened 2025-01-09
Feat: Allow closures to be registered as operation extensions

joelbutcher:
## TL;DR

This PR makes it possible for developers to define a closure as an `OperationExtension` versus having to create class that extends the abstract.

## How it works
I've updated the `Scramble::registerExtension` method to accept both `string` and `callable` scalar types as the `$extension` parameter (renamed as appropriate). We then do a type check inside this method to cast callable as instances of `\Closure` and then register the extension:

```php
$extension = is_string($extension) ? $extension : $extension(...);

static::$extensions = array_merge(static::$extensions, [$extension]);
```

For the `registerExtensions` method, I've simply replaced the `array_merge` with a foreach and a call to `registerExtension`. This is a little bit less performant than previously, but unless an application is registering an ungodly number of extensions, the difference is negligible.

> [!NOTE]
> I've also updated the PHPDoc's on these methods and the `Scramble::$extensions` property.

## Example
To register an operation extension as a closure, add the following code to the `boot` method of your applications `AppServiceProvider`:

```php
Scramble::registerExtension(function(Operation $operation, RouteInfo $routeInfo): void {
    // ...
});
```

If you wish to register multiple extensions as closures, you can do so by calling `Scramble ::registerExtensions(...)`:

```php
Scramble::registerExtensions([
    function(Operation $operation, RouteInfo $routeInfo): void {
        // Extension 1
    },
    function(Operation $operation, RouteInfo $routeInfo): void {
        // Extension 2
    }
    ...
]);
```

### DIFF
--- a/src/Scramble.php
+++ b/src/Scramble.php
@@ -17,6 +17,9 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Route as RouteFacade;
 use LogicException;
 
+/**
+ * @phpstan-type OperationExtensionClosure = Closure(Operation, RouteInfo): void
+ */
 class Scramble
 {
     public static $routeResolver = null;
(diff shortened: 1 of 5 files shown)


joelbutcher:
@romalytvynenko is this still something you want adding to the packages codebase?

PR #710, opened 2025-02-07
"type" should be a single value and "nullable" is available

qooplmao:
I had issues with the generated schema when trying to pull in to a client generator (jane-php) so I had a look around and it turned out the nullable types were causing things to kick off.

[The docs](https://swagger.io/docs/specification/v3_0/data-models/data-types/) say type" should be a single value and "nullable" is available.

### DIFF
--- a/src/Support/Generator/Types/Type.php
+++ b/src/Support/Generator/Types/Type.php
@@ -84,7 +84,8 @@ abstract class Type
     {
         return array_merge(
             array_filter([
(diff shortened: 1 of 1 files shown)


qooplmao:
No worries.

I'm using the project (and think it's excellent, BTW) and have monkey patched an override so I can use the client generator so thought I would contribute back but it seems I'm working with old specs.

I have to say that I think I prefer `nullable: true` over `[string, 'null']` but it is what it is.

Issue #751, opened 2025-03-01
Feature request - optionally expose excluded routes.


willvincent:
### Feature request 🎁

The `ExcludeRouteFromDocs` and `ExcludeAllRoutesFromDocs` attributes are super handy to completely exclude routes, but it would be really useful to be able to flag routes as private or internal, and optionally be able to include those if rendering docs for a user with appropriate permissions.  Even if this were only available as an option on the artisan `scramble:export` command.

Maybe the simplest option is to add an optional flag to the export command that _ignores_ the exclude attributes?

### Reasoning

There are often times where only a portion of the API is intended for public or third party consumption, but it would be beneficial to maintain api documentation for the internal team who has access to everything. ie: a frontend team integrating with a backend api when building out a SPA or mobile app that is not directly served by the backend code (as is my case).

PR #752, opened 2025-03-01
Optionally expose excluded routes

willvincent:
Addresses Issue #751

- Adds an `--all` option to the export command that ignores `ExcludeRouteFromDocs` and `ExcludeAllRoutesFromDocs` attributes
- Also ignores `ExcludeRouteFromDocs` and `ExcludeAllRoutesFromDocs` in the UI if the request has the hiddenContext `scramble-all-access`:

    One need simply set the hidden context for appropriate users' requests, thusly:

    ```php
    Context::addHidden('scramble-all-access', true);
    ```

### DIFF
--- a/src/Console/Commands/ExportDocumentation.php
+++ b/src/Console/Commands/ExportDocumentation.php
@@ -12,6 +12,7 @@ class ExportDocumentation extends Command
     protected $signature = 'scramble:export
         {--path= : The path to save the exported JSON file}
         {--api=default : The API to export a documentation for}
+        {--all : Ignore exclude attributes, export all routes}
     ';
 
     protected $description = 'Export the OpenAPI document to a JSON file.';
@@ -19,6 +20,10 @@ class ExportDocumentation extends Command
     public function handle(Generator $generator): void
     {
         $config = Scramble::getGeneratorConfig($api = $this->option('api'));
+
+        if ($this->option('all')) {
+            $config->set('ignore_exclude_attributes', true);
+        }
 
         $specification = json_encode($generator($config), JSON_PRETTY_PRINT);
 
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -185,8 +185,12 @@ class Generator
             })
             ->filter($config->routes())
             ->filter(fn (Route $r) => $r->getAction('controller'))
-            ->filter(function (Route $route) {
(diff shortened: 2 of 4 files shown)


willvincent:
Probably would be good to be able to optionally ignore hidden properties if that all access flag is enabled too. 

This way if, when using laravel-data for instance with conditional lazy properties that will only be included with appropriate access, they can be hidden from api docs, but optionally rendered for a user with adequate permission.

willvincent:
Yes I have multiple docs, but this is all one API routes don't differ, just some of it is going to be publicly accessible and some won't be - permissions based, _not_ route based. In fact I want to control what's available to users based on permissions at the same routes.

For example an internal user (or our own frontend app) might get 15 properties on a route, that I may only want 4 or 5 to be available to general users.

Some routes will be internal only, but it's not strictly route-related.

Hope that's clear, long day.. ;)

willvincent:
Not really, my use case is more involved than that.. 

I guess I don't understand the push back to having an option to be _able_ to render everything, whether the majority of people would use it or not. 

Even if _only_ for the artisan generate command. 🤷

willvincent:
I would have to explicitly define _many_ routes, to setup yet another doc config for internal or not.

And it still doesn't address my other comment (nor does my PR) about being able to set properties hidden, but then also expose _those_ given appropriate access, or .. again, manually generating docs on the cli.

The ultimate goal here is I'd like to be able to have a full and complete document of _everything_ for my internal team. and by way of excluding routes and hiding properties, provide a curated subset of routes _and_ request and response parameters.

Ideally I'd like to not have to manage a separate doc config which would be.. substantial effort.

I have several hundred routes to wrangle here. 😂

willvincent:
That could work - would be very useful to have available for properties too in request and response.. defined via validation rules, laravel-data properties, manually configured, etc. 

Not sure how that would work if things are defined in comments, er.. I guess we already have the `@ignoreParam` thing, maybe something similar to allow only for a certain api doc? 

I do like the annotation approach, and it's a pattern already in play, so it does make sense to lean into that even more.

romalytvynenko:
Closing the PR due to implementation of `#[Api]` attribute: https://github.com/dedoc/scramble/pull/1211

PR #777, opened 2025-03-18
Make RulesMapper class can be replaced

liquid207:
Allow users of this package to use custom `RulesMapper` class, so the user can add below validation rules to  `RulesMapper` class.

- custom validation rules,
- validation rules not supported by this package

Example:

```
class RulesMapper extends \Dedoc\Scramble\Support\OperationExtensions\RulesExtractor\RulesMapper
{
    public function age(Type $type)
    {
        $integer_type = new IntegerType;
        $integer_type->min = 0;
        $integer_type->max = 18;

        return $integer_type;
    }
}
```


liquid207:
@romalytvynenko 

I am happy to contribute different rules support, but the rules I want support are custom rules (only works for my project), so I will wait you add the support of custom validation rules.

liquid207:
@romalytvynenko 
How are you going to implement this? Can you share your thoughts, so maybe I can submit a PR for this.

PR #811, opened 2025-04-23
[feature] Allow for multiple `@status` responses

MarnuLombard:
- By allowing for parsing more than one `@status` doc-block we can export the same response under one code
- This is especially useful as Laravel Resources will return status code 201 is the underlying model `wasRecentlyCreated`

### DIFF
--- a/src/Support/OperationExtensions/RulesExtractor/RulesToParameter.php
+++ b/src/Support/OperationExtensions/RulesExtractor/RulesToParameter.php
@@ -53,6 +53,13 @@ class RulesToParameter
                 : $this->getTypeFromObjectRule($type, $rule);
         }, new UnknownType);
 
+        if (
+            $rules->contains('sometimes')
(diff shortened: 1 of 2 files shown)


MarnuLombard:
Hi @romalytvynenko 
Any upsert endpoint could return a 200 OR a 201 response based on whether the `Model::$wasRecentlyCreated` is true or not.
In fact that is the [default behaviour of Laravel when converting a resource to a response.](https://github.com/laravel/framework/blob/44e6a294e4441e9e3338008af0288979b3f677e8/src/Illuminate/Http/Resources/Json/ResourceResponse.php#L111-L120)

There is no way (currently) to document that behaviour using scramble.

This code modifies no current behaviour of the library (manual `@status` doc-blocks are still respected), but adds the feature of displaying a response under multiple status codes.

MarnuLombard:
Hi @romalytvynenko 

Will this be merged?

PR #813, opened 2025-04-26
Support response headers

happyDemon:
I was sad to find out that response headers were not yet supported by scramble, so I created this POC PR to implement them.

I say POC, because my implementation feels a bit hack-y. I'm open to any feedback before I add tests.

OpenAPI 3.0 supports response headers: https://swagger.io/docs/specification/v3_0/describing-responses/#response-headers

## Sample case
[Laravel's ratelimiter](https://laravel.com/docs/12.x/routing#rate-limiting) adds the `X-Ratelimit-Limit` and `X-Ratelimit-Remaining` headers to all requests that implement the `throttle` middleware. When the rate limit is hit, the 429 request will also include the `X-Ratelimit-Remaining` header.

## Implementation

On controller-level, a developer can add #[BodyHeader('X-Rate-Limit', 'description', 'type', 'format')] -> I hacked something together so the header(s) get added to the response.

In `withOperationTransformers` it's possible to add/modify responses, adding headers through `registerHeader(string $name, ?string $description, Type $type)`

### DIFF
--- a/src/Attributes/BodyHeader.php
+++ b/src/Attributes/BodyHeader.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Dedoc\Scramble\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_ALL)]
+class BodyHeader extends Parameter
+{
+    public function __construct(
+        string $name,
+        ?string $description = null,
+        ?string $type = null,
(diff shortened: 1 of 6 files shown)


happyDemon:
ah, this PR broke some stuff.
I also did some reading, seems like the BodyHeader parameter approach is not correct, I'll definitely have to rework this.

PR #890, opened 2025-07-04
Improved documentation of complex query parameters (arrays, objects)

### LINKED ISSUE #650: Object field in query parameters doesn't sent correctly to API
I have some query parameters like this:
image

I tried to send filters[search] using the Swagger page as shown here:
image

but it didn't work as expected:
image

Am I doing something wrong, or does this need to be fixed?

### LINKED ISSUE #767: Arrays of values are not properly sent
I'm trying to send an array of strings, the "laravel way", so not in a single query param and then split it, but more like an array of query params, ex

tags[]=a&tags[]=b&tags[]=c etc

Not sure if Scramble supports this out of the box, but from my attempts, I haven't really been able to get it to work, you can find more info below

Environment:
- Laravel 11
- Scramble v0.12.11

Example: Validation rules

```php
public function rules()
    {
        return [
            'first' => 'nullable|integer|min:1|max:100',
           
            'page' => 'nullable|integer|min:1',
    
            'channel' => 'nullable|string|max:16',

            /**
             * Filter by tag slugs
             * 
             * @example ["french", "norwegian"]
             */
            'tags' => 'sometimes|array',

            /**
             * Filter by status
             * 
             * @example ["live", "past"]
             */
            'status' => 'sometimes|array',
        ];
    }
```

Generated CURL Request
```sh
curl --request GET \
  --url 'https://api.....local/api/v2/streams?tags=french' \
  --header 'Accept: application/json' \
```

Response, makes sense, since it's not formatted as an array of strings
```sh
{
  "message": "The tags must be an array.",
  "errors": {
    "tags": [
      "The tags must be an array."
    ]
  }
}
```

Expected CURL Request
```sh
curl --request GET \
  --url 'https://api.streamtest.local/api/v2/streams?tags[]=french' \
  --header 'Accept: application/json' \
```

### LINKED ISSUE #879: [Bug]: Using a custom AllowedFilter in Spatie QueryBuilder results in two parameters in the generated OpenAPI spec
### What happened?

Using a single `AllowedFilter::custom()` in QueryBuilder creates two params in the spec:

```json
"parameters": [
    {
        "name": "filter[accountable]",
        "in": "query",
        "schema": {
            "type": "string"
        }
    },
    {
        "name": "filter",
        "in": "query",
        "schema": {
            "type": "object",
            "properties": {
                "accountable": {
                    "type": "string"
                }
            }
        }
    }
],
```

I believe the first parameter with `"name": "filter[accountable]"` is the correct one?

### How to reproduce the bug

My QueryBuilder looks like this:

```php
$accounts = QueryBuilder::for(Account::class, $request)
    ->allowedFilters([
        AllowedFilter::custom('accountable', new BusinessAccountablesFilter),
    ])
    ->get();
```

### Package Version

0.12.23

### PHP Version

8.3.20

### Laravel Version

11

Linux

### Notes

_No response_

romalytvynenko:
This PR improves documentation of complex query parameters (arrays and objects) so it better suits available OpenAPI tools. Currently OpenAPI specification (3.1) and available tools don't allow to easily document complex query parameters (see https://github.com/OAI/OpenAPI-Specification/issues/1706).

Particularly, this PR adds handling of 2 cases. 

**1. Objects and nested objects (fixes #650)**

Imagine you have the following validation rule for your query parameters:
```php
$request->validate([
    'filter.name' => ['string'],
]);
```

You can then pass some value with the following query string:
```
?filter[name]=something
```

Previously Scramble documented this as `filter` parameter that is an `object` with the `name` property:
```json
{
  "name": "filter", 
  "in": "query", 
  "schema": {
    "type": "object", 
    "properties": {
      "name": {"type": "string"}
    }
  }
}
``` 
This seems okay, but it didn't work properly in `Try it out` section in Stoplight Elements (you've got `name=something` in the example CURL section when provided `{"name":"something"}` for `filter` parameter). Yes, you could've used `deepObject` for `style` in this specific case, but not all existing tools will serialize it correctly even with `deepObject`! For example, Stoplight Elements which Scramble uses by default will not properly serialize it.

With this PR, Scramble will document such query parameters like this:
```json
{"name": "filter[name]", "in": "query", "schema": {"type": "string"}}
```

This makes documentation clear and `Try it out` section work properly in all OpenAPI tools. Also this is how Scramble PRO documents filters for Spatie Query Builder package.

This also fixes #879 by not keeping duplicated query parameters.

**2. Array of scalars (fixes #767)**

This is the case when you have a query parameter defined by the following rules:
```php
$request->validate([
    'tags.*' => ['string']
]);
```

`tags` parameter now can be passed via this query string:
```
?tags[]=foo&tags[]=bar
```

Previously Scramble documented such parameter as `tags` parameter which is an `array` of `string`:
```json
{"name": "tags", "in": "query", "schema": {"type": "array", "items": {"type": "string"}}}
``` 

Again, this wasn't properly handled by existing OpenAPI tooling. Not cool! So now Scramble documents it in a way that is understandable by tools (meaning that they produce a correct query string):
```json
{"name": "tags[]", "in": "query", "schema": {"type": "array", "items": {"type": "string"}}}
```

As you can see, Scramble adds `[]` suffix to the parameter name. You can see the discussion about this in this issue's comments #767.

**What is not handled in this PR**

Even more complex structures for query parameters, such as arrays of objects, etc., are left intact (*almost*). There is simply no good way to document them for the existing OpenAPI tooling. 

Scramble, however, marks them with `"x-deepObject-style": "qs"` ("qs" because `qs` library serializes deep objects in the way that is understood by PHP). It is not used by any tools out there but *it can be used* either by tools, or by a developer who uses Scramble. 

In [operation transformer](https://scramble.dedoc.co/developers/customize-openapi-documents#operation-transformers) you can iterate over parameters and if any parameter has this extension set to `qs` you can do whatever you want. For example, you can add a description helping your users to understand how to properly pass the parameter. Or you can flatten it. Let me know if you do something cool with it!

🥴

### DIFF
--- a/src/Support/OperationExtensions/RequestBodyExtension.php
+++ b/src/Support/OperationExtensions/RequestBodyExtension.php
@@ -173,7 +173,7 @@ class RequestBodyExtension extends OperationExtension
     }
 
     /**
-     * @param Parameter[] $params
+     * @param  Parameter[]  $params
      * @return Parameter[]
      */
     protected function convertDotNamedParamsToQueryParams(array $params): array
@@ -235,8 +235,8 @@ class RequestBodyExtension extends OperationExtension
                     return null;
                 }
 
-                if (Str::endsWith($parameter->name, '[]') && !$parameter->schema->type instanceof ArrayType) {
-                    $parameter->schema->type = (new ArrayType())
+                if (Str::endsWith($parameter->name, '[]') && ! $parameter->schema->type instanceof ArrayType) {
+                    $parameter->schema->type = (new ArrayType)
                         ->setItems($parameter->schema->type)
                         ->addProperties($parameter->schema->type);
                 }
--- a/tests/Support/OperationExtensions/RequestBodyExtensionTest.php
+++ b/tests/Support/OperationExtensions/RequestBodyExtensionTest.php
@@ -604,7 +604,7 @@ class RequestBodyExtensionTest_DeepQueryParametersController
     public function __invoke(Request $request)
     {
         $request->validate([
-            'filter.accountable' => 'integer'
+            'filter.accountable' => 'integer',
         ]);
     }
 }
@@ -647,7 +647,7 @@ it('documents array query parameters as arrays of some type', function () {
                 'type' => 'array',
                 'items' => [
                     'type' => 'string',
-                ]
+                ],
             ],
         ]);
 });
@@ -674,7 +674,7 @@ it('documents array query parameters as arrays of specific type', function () {
                 'type' => 'array',
                 'items' => [
                     'type' => 'integer',
-                ]
+                ],
             ],
         ]);
 });


darkbasic:
This is a HUGE BC for us and breaks the whole FE library that is supposed to validate filter typings.

darkbasic:
@romalytvynenko I'm sorry for the late reply but we had some issues with the new param and we didn't manage to debug them until now.

image

This is how we instantiate scramble. `'flatten_deep_query_parameters' -> false` doesn't work there but if we instead modify the default value in the package itself it does. We ended up overriding the package defaults for the moment until you manage to get it working at instantiation time. Thanks!

PR #923, opened 2025-08-05
add support for automatic color mode

macbookandrew:


macbookandrew:
@romalytvynenko absolutely, that makes more sense; updated. Thanks!

PR #941, opened 2025-08-27
feat(operations): support OperationId via attribute

jrdnrc:
This PR adds the ability to generate an operation ID via an Attribute, similar to some of the other attributes supported :)


jrdnrc:
@romalytvynenko good shout, thanks!

I've also changed `INF` to `PHP_INT_MAX` in the Endpoint attribute as `INF` is a float

Issue #945, opened 2025-09-02
[Bug]: `@var list<T>` does not infer correctly


chrisvanlier2005:
### What happened?

I would expect `list<T>` to be inferred a s `array<int, T>` when serializing to the OpenAPI specification. This is however currently not supported and results in the following json

Expected:
```
"custom_fields": {
    "type": "array",
    "items": {
        "$ref": "#/components/schemas/CustomField"
     },
},
```

Actual:
```
"fields": {
    "type": "object"
},
```

### How to reproduce the bug

1. Create a json resource
2. Add a property with the following phpdoc
```
/**
 * @var list<string>
 */
'some_list' => $this->resource->some_list
```

### Package Version

0.12.29

### PHP Version

8.4

### Laravel Version

12.22.1

### Which operating systems does with happen with?

macOS

PR #961, opened 2025-09-17
fix: Add support for transforming @var list<T>

chrisvanlier2005:
Should resolve: https://github.com/dedoc/scramble/issues/945

Not 100% where the tests should be located.


chrisvanlier2005:
Makes perfect sense to keep the type system minimal! I'll refactor to use `array<int, T>` mapping instead of adding a new type.

PR #996, opened 2025-10-17
Support more integer types

Neol3108:
I often use more strict integer types in my phpdocs where possible, but I've always had to document them as int because of Scramble: https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges

### DIFF
--- a/src/PhpDoc/PhpDocTypeHelper.php
+++ b/src/PhpDoc/PhpDocTypeHelper.php
@@ -9,6 +9,7 @@ use Dedoc\Scramble\Support\Type\FloatType;
 use Dedoc\Scramble\Support\Type\FunctionType;
 use Dedoc\Scramble\Support\Type\Generic;
 use Dedoc\Scramble\Support\Type\GenericClassStringType;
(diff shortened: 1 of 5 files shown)


romalytvynenko:
@Neol3108 thanks, awesome. Can you share the real example where you use this types in the real code? 

I'm interested in seeing the real case

src/PhpDoc/PhpDocTypeHelper.php
```diff
@@ -194,4 +213,53 @@ private static function handleIdentifierNode(IdentifierTypeNode $type)
 
         return new ObjectType($type->name);
     }
+
+    /**
+     * @param  list<\PHPStan\PhpDocParser\Ast\Type\TypeNode>  $genericTypes
+     */
+    private static function handleGenericInteger(array $genericTypes): IntegerType
```
Neol3108: I personally don't like the amount of if statements that are in this method, might look into it later

romalytvynenko:
@Neol3108 thanks! Looking good!

PR #1005, opened 2025-10-25
feat: parse variables used in validation rules, pass http method to request

chaikivskyi:
### Context
In some controllers we reuse the same action for multiple HTTP methods. For example, with:

`Route::apiResource('tests', TestController::class);`

Laravel routes **PUT** `/tests/{id}` and **PATCH** `/tests/{id}` to `TestController@update`.
Functionally, the primary difference between **PUT** and **PATCH** is usually the validation rules:

For **PUT**, all fields are typically required (full replacement).

For **PATCH**, fields typically use sometimes (partial update).

A common pattern is:

```
$requirement = $request->isMethod(Resoinse::METHOD_PUT) ? 'required' : 'sometimes';

$request->validate([
    'test_string_field'  => [$requirement, 'string'],
    'test_numeric_field' => [$requirement, 'integer'],
]);
```
### Problem

When we parse validation rules for rule extraction, we only consider parameters explicitly declared in the method signature and don't set the request method early enough. As a result, method-dependent variables like `$requirement` can’t be resolved correctly, leading to inaccurate rule parsing for **PUT** vs **PATCH**.

### Proposal

1. Define request method before parsing rules.
2. Evaluate variables used inside rule arrays.

**PS.** Please, do not close this PR, lets discuss every hesitance in this thread and I will apply fixes ASAP

### DIFF
--- a/src/Support/OperationExtensions/RulesEvaluator/NodeRulesEvaluator.php
+++ b/src/Support/OperationExtensions/RulesEvaluator/NodeRulesEvaluator.php
@@ -31,10 +31,9 @@ class NodeRulesEvaluator implements RulesEvaluator
             return [];
         }
 
-        $parameters = $this->evaluateParameters();
-        $variables = $this->evaluateVariables($parameters);
+        $vars = $this->evaluateDefinedVars();
 
-        $rules = $this->evaluateExpression($this->rulesNode, array_merge($variables, $parameters)) ?? [];
+        $rules = $this->evaluateExpression($this->rulesNode, $vars) ?? [];
 
         foreach ($rules as &$item) {
             if (is_string($item)) {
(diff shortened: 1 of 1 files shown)


chaikivskyi:
Hi @romalytvynenko 
> Can you please run PHPStan locally and fix issues in the newly added code?

Thank you, fixed

> The only concern I have now is if variable is declared/re-declared in `if` block, etc:
> 
> ```
> $required = false;
> if ($request->isMethod('PUT')) {
>     $required = true;
> }
> ```
> 
> p.s.: But I'm not sure if it would be wise to handle this `if` case issue. Just interested to hear your thoughts on it

Actually, there are many more cases we could cover, but I'd start with the simplest use case and improve this functionality as needed. If many developers request it, then it makes sense to add it. For now, I'd keep it simple.

PR #1012, opened 2025-11-05
Support for classes which implement `Arrayable` interface

klAvAx:
Currently supported models are limited to Eloquent only, if a non-Eloquent model in `App\Models\...` namespace is used then it will not be documented and return generic type for the whole model.

This seems like a fairly simple addition to add support for non-Eloquent models which implement `Arrayable`  on top of the existing Eloquent model support.

This support can be made even more generic by removing line 34 @ `src/Support/TypeToSchemaExtensions/ModelToSchema.php`
Then if Model has a `toArray` function it will be documented regardless if it's Eloquent or completely generic with no implemented interfaces.

Using `stripos` for 2 reasons:
- is case-insensitive.
- works from PHP 5

it can be replaced with `str_contains` if case sensitivity needs to be more strict and/or PHP support will be never less than 8

Any thoughts about this?

### DIFF
--- a/tests/Support/TypeToSchemaExtensions/ArrayableToSchemaTest.php
+++ b/tests/Support/TypeToSchemaExtensions/ArrayableToSchemaTest.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Dedoc\Scramble\Tests\Support\TypeToSchemaExtensions;
+
+use Dedoc\Scramble\GeneratorConfig;
+use Dedoc\Scramble\Infer;
+use Dedoc\Scramble\OpenApiContext;
+use Dedoc\Scramble\Support\Generator\Components;
+use Dedoc\Scramble\Support\Generator\OpenApi;
+use Dedoc\Scramble\Support\Generator\TypeTransformer;
+use Dedoc\Scramble\Support\Type\ObjectType;
+use Dedoc\Scramble\Support\TypeToSchemaExtensions\ArrayableToSchema;
+use Illuminate\Contracts\Support\Arrayable;
+
+beforeEach(function () {
+    $this->components = new Components;
+    $this->context = new OpenApiContext((new OpenApi('3.1.0'))->setComponents($this->components), new GeneratorConfig);
+    $this->transformer = new TypeTransformer(app(Infer::class), $this->context, [
+        ArrayableToSchema::class,
+    ]);
+});
+
+it('transforms arrayable to schema', function () {
+    $schema = $this->transformer->transform(new ObjectType(Foo_ArrayableToSchemaTest::class));
+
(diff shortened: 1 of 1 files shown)


src/Support/TypeToSchemaExtensions/ModelToSchema.php
```diff
@@ -28,7 +29,13 @@ public function __construct(
     public function shouldHandle(Type $type)
     {
         return $type instanceof ObjectType
-            && $type->isInstanceOf(Model::class)
+            && (
+                (
+                    $type->isInstanceOf(Arrayable::class)
+                    && (stripos($type->name, '\\Models\\') !== false)
+                )
```
chrisvanlier2005: In my opinion, I would rather see an `ArrayableToSchema` extension which transforms objects from any namespace. Thoughts?

klAvAx: Hmm, for an `ArrayableToSchema` i would drop the `stripos` check to make it more universal and not limited to `\Models\` namespace path 🤔
What do you think about that?

klAvAx:
@romalytvynenko Sorry figured that i had to do that 😅 

PHPStan baseline rolled back.

For consideration laravel Eloquent models implement `Arrayable`, perhaps the `ModelToSchema` could just be converted to `ArrayableToSchema`?
That would automatically support all classes which implement `Arrayable` regardless if its a Model or Data object or ...

romalytvynenko:
@klAvAx can you also share some real case where you use not model but arrayable in a resource?

klAvAx:
@romalytvynenko Sure, lemme whip up a demo repo to properly showcase things, and yes check change in json resource helper is needed to make it better supported.

klAvAx:
> I don't think this is just for replacing models with `Arrayable` objects. This could give some basic support for serializing custom objects:
> 
> ```diff
> /**
>  * @property \App\Models\User $resource
>  */
> final class UserResource extends JsonResource
> {
>     /**
>      * Transform the resource into an array.
>      *
>      * @param \Illuminate\Http\Request $request
>      * @return array<string, mixed>
>      */
>     public function toArray(Request $request): array
>     {
>         return [
>             'id' => $this->resource->uuid,
>             'name' => $this->resource->name,
> +           /** @var list<\App\Value\CustomField> **/
>             'custom_fields' => $this->resource->custom_fields,
>         ];
>     }
> }
> ```

Yeah, did you test your use-case with this PR?
for my use cases and for the example git repo i provided it works a treat with 1 exception which i think i should tackle at some point to make it a little better, it might not be the perfect implementation but if it works great and does not tank performance i'd say is all good? 😅

PR #1054, opened 2026-01-01
Improce doc sort order

chaikivskyi:
### Problem

When a group does not have a weight defined, the sorting order becomes unpredictable.
As a result, the API documentation appears cluttered and difficult to navigate.

### Solution

Introduce a secondary sorting rule:
Primary: sort by weight (when defined)
Fallback: sort by group name when weight is not set

### Result

This change ensures a consistent and predictable order in the API documentation, making it cleaner, more readable, and significantly easier to find the required groups.

### DIFF
--- a/src/DocumentTransformers/AddDocumentTags.php
+++ b/src/DocumentTransformers/AddDocumentTags.php
@@ -54,8 +54,8 @@ class AddDocumentTags implements DocumentTransformer
 
         return $tags
             ->sort(function (Tag $a, Tag $b) {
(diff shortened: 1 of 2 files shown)


romalytvynenko:
@chaikivskyi aren't endpoints sorted based on their definition order by default?

chaikivskyi:
No, on my project, the first endpoint displayed comes from line 47 of one of the route files.
Sorting endpoints purely by definition order isn't ideal in my case.
If this behavior is useful for other devs, it may be worth adding a configuration option to allow different sorting strategies.

PR #1057, opened 2026-01-02
Add property types from PHPDoc support

chrisvanlier2005:
Still work in progress, need to add some tests. Bit i've opened it early to gauge interest

----

The logic is modularized using a `PropertyExtractor` interface, allowing for distinct strategies to handle different locations where type information might be defined.

Examples

Here are the specific class scenarios that are now supported with this extension:
## Standard Properties with `@var`

Types are now inferred from the `@var` annotation above the property definition.

```php
class UserData
{
    /** @var string */
    public $name;

    /** @var \Illuminate\Support\Collection<int, string> */
    public $tags;
}
```

## Magic Properties (`@property`)

Properties defined in the class DocBlock (common in Laravel Models or fluent objects) are now recognized.

```php
/**
 * @property string $status
 * @property-read int $count
 */
class ServerStats
{
    // ...
}
```

## Constructor Promoted Properties

Types are inferred from the constructor's `@param` tags when using promoted properties.

```php
class CreateUserRequest
{
    /**
     * @param string $email
     * @param int|null $age
     */
    public function __construct(
        public $email,
        public $age,
    ) {}
}
```


romalytvynenko:
@chrisvanlier2005 have you been able to solve this problem on your projects?

chrisvanlier2005:
@romalytvynenko Yes! We've currently implemented this code as a custom extension, which we use in all of our repositories. However recently I haven't got the time to update this PR with tests and kind of forgot about it.

It should be completely functional, except for this debugging statement: https://github.com/dedoc/scramble/pull/1057/changes#diff-85a69a822f8706e01bf932c88a638eb6fbca1080e55258f1ec2531cbb9a9da89R37

romalytvynenko:
Merging this in!

PR #1122, opened 2026-03-27
Infer 201 status code for JsonResource wrapping a newly created model

### LINKED ISSUE #1071: [Bug]: `JsonResource` response for a newly created model documented as `200` instead of `201`
### What happened?

When returning a `JsonResource` for a newly (in that request) created model Laravel returns a `201` but Scramble documents this as a `200` response. I expected this status code to be the correct one.

### How to reproduce the bug

This is documented as a `200` response. In reality this is a `201` response.

```
$entity = new EntityModel;

// ... setting properties, doing other stuf etc.

$otherEntity->relation()->save($entity);

return EntityResource::make($entity);
```

### Package Version

0.13.10

### PHP Version

8.5.2

### Laravel Version

12.48.1

_No response_

### Notes

Not sure this is fixable but otherwise it might be good to document this properly somehow since it can be pretty confusing. Or maybe I just completely missed the warnings about this, but it caught me off guard doing an implementation with my own spec 🤣

dbrekelmans:
Fixes #1071 

When a controller creates a model via Builder::create() / forceCreate() / forceCreateQuietly() and returns it wrapped in a JsonResource, Laravel automatically responds with 201 (because Model::$wasRecentlyCreated is true). Scramble previously always documented this as 200.

This PR adds detection for this pattern to document the correct 201 status code.

```php
// Now correctly documented as 201
public function store(Request $request)
{
    $user = User::create($request->validated());
    return new UserResource($user);
}
```

### How it works
1. EloquentBuilderExtension marks model types returned from create(), forceCreate(), and forceCreateQuietly() with a wasRecentlyCreated type attribute
2. ResourceResponseTypeToSchema checks for this attribute when determining the base response status code, using 201 instead of the hardcoded 200
3. Explicit withResponse() or ->response()->setStatusCode() overrides still take precedence

### Scope
This covers the most common and unambiguous creation patterns — methods that always create a new model. Ambiguous methods like firstOrCreate() and updateOrCreate() (which may or may not create) are intentionally left at 200.

### DIFF
--- a/src/Support/InferExtensions/EloquentBuilderExtension.php
+++ b/src/Support/InferExtensions/EloquentBuilderExtension.php
@@ -20,6 +20,10 @@ class EloquentBuilderExtension implements MethodReturnTypeExtension
 
     public function getMethodReturnType(MethodCallEvent $event): ?Type
     {
+        if ($this->isModelCreationMethod($event->getName())) {
+            return $this->handleModelCreation($event);
+        }
+
         if ($event->getDefinition()->hasMethodDefinition($event->getName())) {
             return null;
         }
@@ -102,6 +106,23 @@ class EloquentBuilderExtension implements MethodReturnTypeExtension
         }
 
         return new Generic($type->name, []);
+    }
+
+    private function isModelCreationMethod(string $name): bool
+    {
+        return in_array($name, ['create', 'forceCreate', 'forceCreateQuietly'], true);
+    }
+
+    private function handleModelCreation(MethodCallEvent $event): ?Type
+    {
+        if (! $modelType = $this->getModelType($event->getInstance())) {
+            return null;
+        }
+
+        $modelType = clone $modelType;
+        $modelType->setAttribute('wasRecentlyCreated', true);
+
+        return $modelType;
     }
 
     private function getScopeMethodName(string $scope): string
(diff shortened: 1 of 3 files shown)


romalytvynenko:
Hey @dbrekelmans 

Was this PR created by AI?

dbrekelmans:
Hey @romalytvynenko, yes I used Claude Code (I am a real human though). I'm checking the output and making sure there are tests covering the change. The first issue I opened a PR for was something I ran into myself when using Scramble. I figured I might as well see if I can quickly fix a few open issues as well to help out, since I like the tool :). No malicious intent and no automated bot, just trying to help you fix a few open issues.

romalytvynenko:
@dbrekelmans gotcha

You see, the reason this issue hasn't been fixed yet is the complexity and the potential architectural changes required to track the state of an object's property (in this case, `wasRecentlyCreated`).

Your PR covers three methods. However, `wasRecentlyCreated` is not only set after running those three methods. Code like `new User()->save()`, `new User()->fill([...])->save()`, or `new User()->forceFill([...])->save()` will also produce a model where `wasRecentlyCreated` is `true`.

There are also different ways to make it `false`, such as `new User()->fill(...)->save()->fresh()` and others.

Long story short, this is why I tend not to accept AI-generated PRs: things may work on the surface, but the depth of edge cases and overall robustness is often lacking.

dbrekelmans:
Hi @romalytvynenko, I didn't realize the depth of this issue, sorry for wasting your time on this one! I'll take more care in the future before I open a PR 🙏

PR #1123, opened 2026-03-27
Fix `Rule::enum` on array items replacing array type in OpenAPI spec

dbrekelmans:
Follow-up to https://github.com/dedoc/scramble/pull/1119#issuecomment-2920547498, applies the same fix to `Rule::enum`.

When `Rule::enum` is used on array items (e.g. `['array', Rule::enum(StatusEnum::class)]`), the EnumRule transformer ignores the previous type and returns the enum type directly, discarding the array entirely.

Before (incorrect):
```json
  {
    "$ref": "#/components/schemas/StatusEnum"
  }
```

After (correct):
```json
  {
    "type": "array",
    "items": {
      "$ref": "#/components/schemas/StatusEnum"
    }
  }
```

### Note on test skips
The `only()` and `except()` array tests use `->skip(! method_exists(Enum::class, 'only'))` / `->skip(! method_exists(Enum::class, 'except'))`, because these methods were only added in Laravel 11. This matches the existing skip conditions on the non-array only/except tests.


romalytvynenko:
@dbrekelmans Thanks!

PR #1124, opened 2026-03-28
Fallback to unit enum name if not backend enum

GJSBRT:


romalytvynenko:
@GJSBRT not sure is this is still relevant to you.

Current implementation will confidently document all not backed enums as unions of their cases names. But this is not correct. Imagine you return such enum from a controller as is:

```php
// app/Enums/Status.php
enum Status
{
    case Draft;
    case Published;
}

// routes/api.php
use App\Enums\Status;
use Illuminate\Support\Facades\Route;

Route::get('/status', function () {
    return response()->json([
        'status' => Status::Draft,
    ]);
});
```

This does not return:

```json
{
  "status": "Draft"
}
```

Instead, JSON encoding fails with:

```text
Non-backed enums have no default serialization
```

It works only when the application converts the enum to its name manually (`$status->name`), or when Laravel serializes an Eloquent enum-cast model attribute.

Because Scramble cannot know which of these cases applies, I do not think we should document every non-backed enum as a string enum by default - this approach can be rather moved to model extension.

I'm keeping this PR open to implement it once I have some time.

PR #1176, opened 2026-06-01
Add IgnoreParam attribute support

WendellAdriel:
## Overview

Adds an `IgnoreParam` PHP attribute that can hide inferred request parameters from generated OpenAPI docs. It works on controller actions, invokable controllers, route closures, and FormRequest classes.

## Why

Scramble already supports `@ignoreParam` in PHPDoc, but attributes are a better fit for projects that prefer native PHP metadata. This keeps the existing annotation behavior while providing the same practical control through an attribute.

## Examples

```php
#[IgnoreParam('secret')]
public function store(Request $request) {}
```

```php
#[IgnoreParam('secret')]
class StoreUserRequest extends FormRequest {}
```

### DIFF
--- a/src/Attributes/IgnoreParam.php
+++ b/src/Attributes/IgnoreParam.php
@@ -8,7 +8,7 @@ use Attribute;
 class IgnoreParam
 {
     /**
-     * @param  'query'|'body'|null  $in
+     * @param  'query'|'path'|'header'|'cookie'|'body'|null  $in
      */
     public function __construct(
         public readonly string $name,
--- a/src/Attributes/IgnoreParam.php
+++ b/src/Attributes/IgnoreParam.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Dedoc\Scramble\Attributes;
+
+use Attribute;
+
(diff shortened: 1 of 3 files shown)


src/Attributes/IgnoreParam.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Dedoc\Scramble\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::IS_REPEATABLE | Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::TARGET_FUNCTION)]
+class IgnoreParam
+{
+    /**
+     * @param  'query'|'body'|null  $in
```
romalytvynenko: @WendellAdriel I would widen this type to all available `in` variants so static analysis tools won't report here.
```suggestion
     * @param  'query'|'path'|'header'|'cookie'|'body'|null  $in
```

romalytvynenko (approved):
@WendellAdriel Looking good! One small change so static analysis tools won't report an error.

Thanks!

PR #1182, opened 2026-06-14
fix: regenerate stale class definitions dictionary (uninitialized $visibility)

hamidreza2005:
## Problem

Analyzing routes that return plain objects fails on a fresh build:

> Typed property `Dedoc\Scramble\Infer\Definition\ClassPropertyDefinition::$visibility` must not be accessed before initialization
> — `src/Support/TypeToSchemaExtensions/PlainObjectToSchema.php:112`

`scramble:clear` + `scramble:export` and `scramble:cache` all throw, so it is unrelated to the document cache from #1178.

## Root cause

`ScrambleServiceProvider` unserializes the shipped `dictionaries/classMap.php` into the `Index` on every boot. That dictionary was generated **before** `$visibility`, `$isStatic`, `$attributes` and `$declaringFileName` were added to `ClassPropertyDefinition`, so its blobs only contain two properties:

...ClassPropertyDefinition":2:{s:4:"type";...s:11:"defaultType";N;}

`unserialize()` does not call the constructor, so the promoted `readonly` defaults are never applied and `$visibility` stays uninitialized. The first access — `PlainObjectToSchema.php:112` — then fatals.

## Fix

Regenerate `dictionaries/classMap.php` via `dictionaries/scripts/generate.php`. The blobs now carry every property (`ClassPropertyDefinition":8`, `visibility` present) and deserialize into complete objects.

## Test

Added `tests/Infer/Definition/ClassPropertyDefinitionTest.php`, which unserializes every shipped blob and asserts each property is accessible. It fails against the stale dictionary and passes after regeneration.


romalytvynenko:
Problem is valid, solution requires more work (file names should not be serialized). I'm working on a fix.

romalytvynenko:
@hamidreza2005 fixed in `0.13.28`

hamidreza2005:
@romalytvynenko 
Awesome, thanks for handling this so quickly, Roma

PR #1191, opened 2026-06-24
fix: preserve validation chain properties (like nullable) on enums

### LINKED ISSUE #1159: [Bug]: `nullable` rule loses nullability when composed with `new Enum(...)` validation rule
### What happened?

When a FormRequest rule combines `nullable` with `new Enum(EnumClass::class)`:

```php
public function rules(): array
{
    return [
        'status' => ['nullable', new Enum(StatusEnum::class)],
    ];
}
```

I expect the generated OpenAPI schema to be both nullable and typed as the enum (`{ "type": ["string", "null"], "enum": [...] }`). Instead Scramble drops the `nullable` flag and only emits the enum type.

### How to reproduce the bug

```php
enum StatusEnum: string
{
    case OPEN = 'open';
    case CLOSED = 'closed';
}
```

```php
class UpdateRequest extends FormRequest
{
    public function rules(): array
    {
        return ['status' => ['nullable', new Enum(StatusEnum::class)]];
    }
}
```

Wire a controller using `UpdateRequest`, generate the OpenAPI spec, and inspect the `status` parameter.

- **Expected**: `{ "type": ["string", "null"], "enum": ["open", "closed"] }`
- **Actual**: `{ "type": "string", "enum": ["open", "closed"] }` — nullable flag missing

### Package Version

0.13.22

### PHP Version

8.4.18

### Laravel Version

12.57.0

_No response_

### Notes

_No response_

hamidreza2005:
Fixes #1159 

This PR fixes a bug where parameters validated using the Enum rule (e.g. new Enum(UserStatus::class)) would lose prior properties in the validation rules chain, such as nullable or default.

For example, a validation rule definition like:

```php
'status' => ['nullable', new Enum(UserStatus::class)]
```
would generate a schema without `nullable: true` in the final OpenAPI documentation.

### Why this happens
During schema extraction, the validation rules are processed sequentially. The `nullable` rule runs first and sets `$previous->nullable = true` . When the `Enum` rule runs, `EnumRule` discarded the `$previous` type state, returning a fresh OpenAPI type that lost all previously-applied attributes.

### DIFF
--- a/src/RuleTransformers/EnumRule.php
+++ b/src/RuleTransformers/EnumRule.php
@@ -18,7 +18,8 @@ class EnumRule implements RuleTransformer
 {
     public function __construct(
         protected TypeTransformer $openApiTransformer,
-    ) {}
+    ) {
+    }
 
     public function shouldHandle(NormalizedRule $rule): bool
     {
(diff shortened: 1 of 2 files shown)


romalytvynenko:
Feels like solution is AI generated.

PR #1197, opened 2026-06-29
feat: add optional hierarchical endpoint grouping

Eng-MMustafa:
﻿> Draft — opened early to get maintainer feedback on scope and direction.

## Motivation
Scramble currently groups endpoints one level deep, via `#[Group]` and tags. As an API grows, that flat list becomes hard to navigate, and there is currently no way to express nested groups (e.g. `Admin → Users`). This adds optional nesting as a natural extension of the grouping concept Scramble already exposes.

## Summary
When enabled, each operation is assigned to a (possibly nested) group, resolved in priority order from the `#[Group]` attribute, configuration rules, the route prefix/name, and finally the controller name. The result is emitted as standard tag `parent` relationships plus the `x-tagGroups` (Redoc-compatible) and `x-tree` document extensions. No UI is included in this PR.

## Features
- Nested endpoint groups (unlimited depth)
- Attribute-based grouping via `#[Group(parent: ...)]`
- Config-based grouping rules (middleware / prefix / namespace)
- Automatic grouping from route prefix or name, with a controller-name fallback

## Backward compatibility
- Disabled by default (`scramble.groups.enabled = false`).
- `parent` is added only as an optional, last parameter on `#[Group]` and `Tag`; no existing call sites change.
- With the feature disabled, generated documents are byte-identical to current output (verified against the full suite).

## Maintainability
Responsibilities are isolated: a strategy pipeline resolves group paths, a self-contained builder assembles the tree, and a thin document transformer wires them together. No duplicated logic, no oversized classes.

## Testing (executed locally)
- Pest: 21 new tests covering attribute, config, route-prefix, middleware and resource-controller grouping, tree building, and the disabled path — all pass.
- Full suite: no regressions versus the base branch.
- PHPStan (level 9) and Pint: clean on all files in this PR.

## Note to maintainers
Feedback is very welcome — I am happy to adjust the public API (e.g. promoting the resolver interface to a contract, or making grouping per-`registerApi()`).

> A follow-up Pull Request containing an optional documentation explorer UI has been prepared separately but is intentionally excluded from this review to keep the scope focused.

PR #1198, opened 2026-06-29
feat: hierarchical documentation explorer (UI for nested groups)

Eng-MMustafa:
﻿> Draft — follow-up to #1197 (grouping infrastructure). Best reviewed after that PR, since this only renders data #1197 produces.

## Summary
Adds an optional, collapsible explorer to the docs view. It renders the `x-tree` extension produced by the grouping infrastructure (#1197) as a nested sidebar over Stoplight Elements. If `x-tree` is absent, the page renders exactly as it does today.

## Scope
A single file: `resources/views/docs.blade.php` (CSS + vanilla JS, no build step, no new dependencies).

## Features (user-visible)
- Collapsible nested group tree
- Sidebar search with match highlighting
- Active-route sync and auto-expand, plus expand/collapse all
- Keyboard focus states, dark-mode aware, mobile drawer
- Persists expand/scroll state in `localStorage`

## Backward compatibility
- Renders only when `x-tree` is present; otherwise the default Elements page is unchanged (the existing system dark-mode handling is preserved).
- No PHP or behavioral changes, and no new dependencies.

## Testing
Behavior verified manually. Output is HTML-escaped (no raw injection of summaries into the DOM). No automated tests are added, as this is a presentation-only view.

## Note to maintainers
This is intentionally a separate PR so the data layer in #1197 can be evaluated on its own. If you would prefer to keep the package UI-light, this can be deferred, reworked, or shipped as a community package — the infrastructure PR stands alone.

### DIFF
--- a/resources/views/docs.blade.php
+++ b/resources/views/docs.blade.php
@@ -1,4 +1,5 @@
-<!doctype html>
+@php($hasExplorer = ! empty($spec['x-tree'] ?? []))
+    <!doctype html>
 <html lang="en" data-theme="{{ $config->renderer()->get('theme', 'light') }}">
 <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


glow2590:
Thank you so much !! @Eng-MMustafa

romalytvynenko:
@Eng-MMustafa is it AI generated?

romalytvynenko:
I don't accept AI PRs (nor I'm ready to accept such large PR without prior discussion).

PR #1200, opened 2026-07-04
query parameter serialization metadata

cappuc:
## Summary

Adds query parameter serialization metadata support to generated OpenAPI Parameter Objects [spec](https://swagger.io/docs/specification/v3_0/serialization/#query-parameters).

Some openapi client generators (like hey-api or orval) uses this parameters to generate the client code to correctly serialize query parameters. 
This is particulary important when not using flattened query parameters because the default `style` is `form` but should be `deepObject` to correctly handle object params (this is an [issue](https://github.com/hey-api/hey-api/issues/3747) I encountered while testing hey-api).

Whit this changes the `style` is always set to `deepObject` for object parameters. I also added a new config `query_array_parameter_explode` that allows to globally set the preferred format for arrays (repeated fields or comma separated).

This allows Scramble to emit standard OpenAPI fields for query parameters:

- `style`
- `explode`
- `allowReserved`

The new API is exposed through `#[QueryParameter]` and a query-specific `QueryParameterStyle` enum.

## Changes

- Added `Dedoc\Scramble\Attributes\QueryParameterStyle` with supported query styles:
  - `form`
  - `spaceDelimited`
  - `pipeDelimited`
  - `deepObject`
- Extended `#[QueryParameter]` with optional serialization arguments:
  - `style`
  - `explode`
  - `allowReserved`
- Added support for serializing `style`, `explode`, and `allowReserved` in OpenAPI parameters.
- Added automatic `deepObject` metadata when `flatten_deep_query_parameters` is disabled and query parameters are documented as object parameters.
- Added `query_array_parameter_explode` config option to document array query parameters as `style: form`, `explode: false` globally.
- Preserved explicit `explode: false` and `allowReserved: false` values in serialized output.
- Kept existing behavior unchanged unless serialization metadata is explicitly configured or inferred from the new config/deep-object behavior.

## Notes

When `flatten_deep_query_parameters` is enabled, object query parameters continue to be flattened and `deepObject` serialization is not forced onto the flattened parameters.

### DIFF
--- a/config/scramble.php
+++ b/config/scramble.php
@@ -140,6 +140,14 @@ return [
      */
     'flatten_deep_query_parameters' => true,
 
+    /**
+     * Determines the default documented array query parameter serialization.
+     *
+     * When `query_array_parameter_explode` is `true`, Scramble relies on OpenAPI defaults (`style: form` and `explode: true`).
+     * When `query_array_parameter_explode` is `false`, Scramble will document array query parameters with `style: form` and `explode: false`.
+     */
+    'query_array_parameter_explode' => true,
+
     'middleware' => [
         'web',
         RestrictedDocsAccess::class,
--- a/src/Attributes/QueryParameter.php
+++ b/src/Attributes/QueryParameter.php
@@ -18,6 +18,9 @@ class QueryParameter extends Parameter
         mixed $default = new MissingValue,
         mixed $example = new MissingValue,
         array $examples = [],
+        public readonly ?QueryParameterStyle $style = null,
+        public readonly ?bool $explode = null,
+        public readonly ?bool $allowReserved = null,
     ) {
         parent::__construct('query', $name, $description, $required, $deprecated, $type, $format, $infer, $default, $example, $examples);
     }
--- a/src/Attributes/QueryParameterStyle.php
+++ b/src/Attributes/QueryParameterStyle.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Dedoc\Scramble\Attributes;
+
+enum QueryParameterStyle: string
+{
+    case Form = 'form';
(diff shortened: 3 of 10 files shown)


romalytvynenko:
@cappuc isn't the underlaying problem is mostly solvable by creating an Operation transformer which changes query parameter's fields?

Also I have not fully understood the initial issue, and I'm very cautious about changing defaults

cappuc:
The openapi specification defines some standard query serialization options that are controlled with the `style` and `explode` fields: https://swagger.io/docs/specification/v3_0/serialization/#query-parameters
The default values when unspecified are `style:form` and `explode:true` which are the most common case but doesn't work well with object (In the spec there are examples).

Some openapi client generators uses that fields to change how query parameters are serialized.
I'm using scramble with `flatten_deep_query_parameters` set to `false` and the main issue I encountered is with clients that defaults to `style: form`  when `style` and this broke the serialization.

My issue can be fixed with just setting `style: deepObject` on object parameters and this is the only thing that changes by default. 
The other changes allows user to override the generated `style` and `object` for a specific parameter and the `query_array_parameter_explode` allow to set `explode: false` for array fields which allows comma separated value  (this can be useful with `spatie/laravel-query-builder` which support that syntax)

I never used the Operation transformer but I will look at it

romalytvynenko:
Thanks for the contribution, but I'm not ready to accept this PR due to another config introduced for a thing that is achievable via an operation transformer.

What I can accept is `style` (string, not enum for now), `explode` on all Parameter* attributes, `allowReserved` on Parameter, but I'm not sure even this is needed taking into account that you can enforce your default desired settings via an operation transformer.

cappuc:
No problem, I implemented the change I needed with an operation transformer

PR #1254, opened 2026-08-20
Add config for optional use of response return comments as response descriptions

yugarinn:
Hello there! First of all, thank you for maintaining this package :)

I am opening this PR to cover for a case that bit me this morning. I needed to temporarily comment out a line to push a hotfix into production, and our pipeline kept regenerating the API docs with the comment:

This was the (bad, I know) hotfix:
```php
        $result->update(['updated_at' => now()]);
        // TODO: The platform squad is working on a fix to add PK columns to
        // the `users_devices` table. Until then we cannot do this update since
        // it would result in a full table update (which is bad).
        // 
        // $result->update(['updated_at' => now()]);

        return new UserDeviceResource($result);
 ``` 
 
 And this is what ended up being generated in our API documentation:
 ```diff
 -                        "description": "`UserDeviceResource`",
+                        "description": "TODO: The platform squad is working on a fix to add PK columns to\nthe `users_devices` table. Until then we cannot do this update since\nit would result in a full table update (which is bad).\n\n$result->update(['updated_at' => now()]);\n\n`UserDeviceResource`",
                         "content": {
                             "application/json": {
                                 "schema": {
```

I ended up moving the commet way above to avoid this, but it would have been great to have the option to ignore it entirely... thus this PR.

Please let me know if this approach seems correct to you, whether or not it is properly tested, or even if you want this at all! This could also be generalized to enable/disable other features. Another approach could be just ignoring comments that are separated from the `return` statement by a newline, but I digress.


romalytvynenko:
@yugarinn Thanks for the PR.  

I will consider making this a default behavior and parse only status and body annotation. I just need to think a bit on it before merging the PR.

romalytvynenko:
Hey @yugarinn,

Instead of adding a new config, I suggest making this the default behavior. A comment should be treated as response documentation only when it uses at least one inline response annotation: `@status`, `@body`, or the new `@description`.

So this will be used as documentation:

```php
/**
 * User profile.
 *
 * @status 201
 */
return new UserResource($user);
```

And this won't be anymore:

```php
/**
 * TODO: Remove after upgrading Laravel.
 */
return new UserResource($user);
```

The new `@description` annotation allows users to provide a response description without being forced to specify `@status` or `@body` just to opt in:

```php
/** @description User profile. */
return new UserResource($user);
```

If `@status` or `@body` is present, things continue to work as they do now, and `@description` isn't required, the docblock prose is used as the description:

```php
/**
 * User created.
 *
 * @status 201
 * @body array{id: int}
 */
return somethingDynamic();
```

Any thoughts on this before I merge?

yugarinn:
Your changes feel like the sanest defaults to be honest, so no, no comments on my end :)

Thank you for looking into this!

PR #1275, opened 2026-09-18
Improve Boost integration

romalytvynenko:
This is the second iteration of Boost integration, with export and analyze commands updates so agents can use them better.


src/Console/Commands/ExportDocumentation.php
```diff
@@ -2,51 +2,92 @@
...
+        $issuesMessage = fn ($summary) => "{$message} with {$summary}.";
 
-        $successMessage = "OpenAPI document exported to {$filename}.";
-        $issuesMessage = fn ($summary) => "OpenAPI document exported to {$filename} with {$summary}. {$verboseSuffix}";
+        $this->renderDiagnostics($result, $successMessage, $issuesMessage);
 
-        if ($this->getOutput()->isVerbose()) {
-            $this->renderDiagnostics($result, $successMessage, $issuesMessage);
-        } else {
-            $this->renderDiagnosticsSummary($result, $successMessage, $issuesMessage);
```
stayallive, line 49: This change breaks when `nunomaduro/collision` is not a production requirement.

In my installation I do not install `nunomaduro/collision` because it's a dev only for my packages and I install with --no-dev.

But since diagnostics are now always rendered scramble:export is unable to run without `nunomaduro/collision` present.

This should either be marked in the dependencies of Scramble or it needs to check and degrade if it's not possible. Or have a flag to ignore diagnostics in a scramble:export which can also be a solution. Maybe a combination of these even.

romalytvynenko: @stayallive thanks for highlighting it. Fixing it

romalytvynenko: @stayallive fixed the issue in https://github.com/dedoc/scramble/releases/tag/v0.13.45

